PR #61, opened 2015-05-28
Update german.txt

danidoedel:


data/language/german.txt
```diff
@@ -2720,24 +2720,24 @@ STR_2714    :-
...
-STR_2727    :???
-STR_2728    :???
-STR_2729    :???
-STR_2730    :???
-STR_2731    :???
-STR_2732    :???
-STR_2733    :???
-STR_2734    :???
-STR_2735    :???
+STR_2718    :(hoch)
```
Gymnasiast: The original string meant 'go up one directory'. Is this the correct German translation for that?

data/language/german.txt
```diff
@@ -2753,22 +2753,22 @@ STR_2747    :{ENDQUOTES}
...
 # New strings used in the cheats window previously these were ???
-STR_2750    :???
-STR_2751    :???
+STR_2750    :Alle Gegenstände nach oben bewegen
+STR_2751    :Alle Gegenstände nach unten bewegen
 STR_2752    :Gras beseitigen
 STR_2753    :Rasen mähen
 STR_2754    :Pflanzen gießen
-STR_2755    :Vandalismus beseitigen
+STR_2755    :{SMALLFONT}Vandalismus beseitigen
```
Gymnasiast: Should not contain {SMALLFONT}

data/language/german.txt
```diff
@@ -2782,21 +2782,21 @@ STR_2775    :Vollbild (Desktop)
 STR_2776    :Sprache
 STR_2777    :{MOVE_X}{SMALLFONT}{STRING}
 STR_2778    :{RIGHTGUILLEMET}{MOVE_X}{SMALLFONT}{STRING}
-STR_2779    :???
-STR_2780    :???
+STR_2779    :Ansichtsfenster #{COMMA16}
```
Gymnasiast: Will shows up 'Ansichtsfenster #1' and such, which I doubt is valid in German. You should probably drop the hash (#).

Gymnasiast:
Thanks for giving the German translation a much needed improvement.
It's looking good, but I found three strings you should take a look at. I put them as line notes.

data/language/german.txt
danidoedel: Yes, the word "Hoch" is also used by Windows Explorer in this context, but I changed it to upper case.
hoch

data/language/german.txt
danidoedel: Ok I fixed that.

data/language/german.txt
danidoedel: You're right, I fixed that.

danidoedel:
I fixed these three strings. 
No Problem, I'm glad to help with the translation.

Gymnasiast:
Thanks! The changes will be synchronised at 05:00, so they'll be in the game tomorrow.

PR #62, opened 2015-05-29
Update german.txt

danidoedel:
Some additional changes and improvements to the German translation.


Gymnasiast:
Looking good again.

But do you know the difference between the two powered launch modes?
Passing station is how the Turbine in Six Flags Belgium works
Not passing station is how the Shuttle Loop from RCT1 worked.

Are your translations correct representations of those?

danidoedel:
Should be more clear and correct now.

PR #69, opened 2015-05-30
Patch 8

danidoedel:
What do you think, shouldn't there be a colon after string 2700, 2738 and 2776? I think it would look more consistent.


Gymnasiast:
Yes, I agree there should be one after string 2700, 2738 and 2776.

danidoedel:
Ok, I'll add them here.

danidoedel:
You can merge if all is fine.
Will you add the colons to UK?

Gymnasiast:
Would you change (..) back to (Hoch), I don't think (..) is easy for non-tech savvy users.

danidoedel:
Ok, saw that in the polish.txt and thought it would be a good idea. I'll change it back.

Gymnasiast:
Thanks!
English UK is already pull requested: https://github.com/IntelOrca/OpenRCT2/pull/1173

Issue #70, opened 2015-05-30
New and Old fields requiring translation (colour schemes)


trigger-death:
There's a lot of new strings in https://github.com/IntelOrca/OpenRCT2/pull/1174

This is a big one since every colourable window now needs a string.

Renamed:
```
STR_5153    :Colour schemes...
```

New:
```
STR_5187    :Finances
STR_5188    :New Campaign
STR_5189    :Research
STR_5190    :Map
STR_5191    :Viewport
STR_5192    :Recent News
STR_5193    :Land
STR_5194    :Water
STR_5195    :Clear Scenery
STR_5196    :Land Rights
STR_5197    :Scenery
STR_5198    :Footpath
STR_5199    :Ride Construction
STR_5200    :Track Design Place
STR_5201    :New Ride
STR_5202    :Track Design Selection
STR_5203    :Ride
STR_5204    :Ride List
STR_5205    :Guest
STR_5206    :Guest List
STR_5207    :Staff
STR_5208    :Staff List
STR_5209    :Banner
STR_5210    :Object Selection
STR_5211    :Invention List
STR_5212    :Scenario Options
STR_5213    :Objective Options
STR_5214    :Map Generation
STR_5215    :Track Design Manager
STR_5216    :Track Design Manager List
STR_5217    :Cheats
STR_5218    :Colour Schemes
STR_5219    :Options
STR_5220    :Keyboard Shortcuts
STR_5221    :Change Keyboard Shortcut
STR_5222    :Load/Save
STR_5223    :Save Prompt
STR_5224    :Demolish Ride Prompt
STR_5225    :Fire Staff Prompt
STR_5226    :Track Delete Prompt
STR_5227    :Save Overwrite Prompt
STR_5228    :{SMALLFONT}{BLACK}Main UI
STR_5229    :{SMALLFONT}{BLACK}Park
STR_5230    :{SMALLFONT}{BLACK}Tools
STR_5231    :{SMALLFONT}{BLACK}Rides and Peeps
STR_5232    :{SMALLFONT}{BLACK}Editors
STR_5233    :{SMALLFONT}{BLACK}Miscellaneous
STR_5234    :{SMALLFONT}{BLACK}Prompts
STR_5235    :{SMALLFONT}{BLACK}Settings
STR_5236    :Window:
STR_5237    :Palette:
STR_5238    :Colour Schemes:
STR_5239    :Duplicate
STR_5240    :Enter a name for the colour scheme
[...]

PR #71, opened 2015-05-30
New strings added, issue #70

danidoedel:


data/language/german.txt
```diff
@@ -2688,12 +2688,12 @@ STR_2681    :{MEDIUMFONT}{BLACK}Erhöht Ihr Geld um 5.000
 STR_2682    :{MEDIUMFONT}{BLACK}Zw. freiem und kostenpfl. Eintritt umsch.
 STR_2683    :{MEDIUMFONT}{BLACK}Erhöht das Vergnügen der Besucher maximal
 STR_2684    :{MEDIUMFONT}{BLACK}Eine große Gruppe von Besuchern tritt ein
-STR_2685    :Simplex Noise Parameters
+STR_2685    :Simplex Noise Parameter
```
Gymnasiast: This is a compound word (Kompositum), which are written without spaces in German, if I'm not mistaken.

Gymnasiast:
Thanks again. I put one remark as a line note.

danidoedel:
Hmm yes and no... ;) 
"Simplex Noise" is itself already a term (http://de.wikipedia.org/wiki/Simplex_Noise), but I think " 'Simplex Noise' Parameter " is a good deal. I'll add it like this...

PR #75, opened 2015-06-01
Theme update, new strings added

danidoedel:


Gymnasiast:
Would you please not touch Dutch again? It's active enough not to be filled with English. I'll take care of it.

PR #133, opened 2015-06-25
Create chinese_traditional.txt

daihakken:
Some strings were changed into official translation. However, it seems that OpenRCT2 is not supporting chinese characters yet, here're screenshots for proving that:
https://farm1.staticflickr.com/345/19138198232_2bcd16befa_o.png
https://farm1.staticflickr.com/442/19147391151_f9da5fa143_o.png

And a screenshot from the vanilla game in Chinese:
https://farm1.staticflickr.com/549/18521415254_acfa295e82_o.png

(P.S. The Chinese font the vanilla game used is PMingLiU, a Chinese "Times New Roman")


Gymnasiast:
OpenRCT2 won't support non-Latin languages for some time. Even those that did work in vanilla RCT2. (Support for those languages was added to RCT2 in not exactly a standard way.)
So I'm afraid merging this will have to wait.

daihakken:
It's okay, thank you for your explanation. This merge is simply a test with Chinese characters, so it's not a big deal at all.

Gymnasiast:
The plan is to move to Unicode (which can only be done after the game is fully decompiled), at which point we can support Traditional Chinese and many other non-Latin languages.

IntelOrca:
You can still create and merge these translations. They will be ready and work when non-latin is supported.

PR #201, opened 2015-08-12
Create korean.txt

telk5093:
Not perfectly translated since it's too late in Korea.
Am I doing right? I'm first to use Github.


Gymnasiast:
Yes, you opened your pull request in the correct way.

PR #208, opened 2015-08-16
beginning of translation

Soosisya:
just first try of translation, so people can see it's not a joke)


Gymnasiast:
You'll have to include all strings up to and including 3446 (even if that's just copies of UK English). Otherwise your translation won't work.

PR #242, opened 2015-09-11
Update russian.txt

Soosisya:
more translating stuff

### DIFF
--- a/data/language/russian.txt
+++ b/data/language/russian.txt
@@ -555,17 +555,17 @@ STR_0550    :
 STR_0551    :
 STR_0552    :
 STR_0553    :
-STR_0554    :The car is accelerated out of the station along a long level track using Linear Induction Motors, then heads straight up a vertical spike of track, freefalling back down to return to the station
-STR_0555    :Guests ride in an elevator up or down a vertical tower to get from one level to another
-STR_0556    :Extra-wide cars descend completely vertical sloped track for the ultimate freefall roller coaster experience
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Thanks. :)
A question: which TTF font works well with Russian? Does Arial work, for example? We need to choose a TTF font in OpenRCT2 to display Cyrillic script.

Soosisya:
yes, it does, Times new roman works as well
By the way, i've downladed the lastest build and still find that russian language isn't listed in the option menu yet

Gymnasiast:
That's right, but I explained above why that was. We will add support for it soon.

IntelOrca:
It would better to use a sans-serif font rather than a serif font if possible.

Gymnasiast:
Well, Arial, then.

IntelOrca:
Yes, we will need be able specify multiple fonts anyway as not all languages are going to have a single font that everyone has. We will also need to allow user override.

tsunamistate:
Arial would be nice, but I also think, Open Sans or Roboto would also be good alternatives, and they're included in most popular Linux distributions.
If you guys would tell how to temporarily include Russian support w/ a font I could experiment and pick a font that looks good.

IntelOrca:
They are defined in language.cpp, you will need to be able to compile the project.

tsunamistate:
Let's continue this conversation on Localisation gitter

PR #254, opened 2015-10-03
Update localisations for wall-clock time based autosaving

Overv:
Translations are based on strings 1879 to 1885, so they should all be correct assuming these were already correct. I've had Dutch, German and French native speakers check the strings.


Gymnasiast:
Would you take out all non-checked languages? We prefer actual translations, not guesses or Google Translated strings.

Overv:
They are not guesses. They were copied from 1879 to 1885, which are the exact same source strings, except with different minute counts.

    STR_1879    :Every 10 minutes
    STR_1880    :Every 20 minutes
    STR_1881    :Every 30 minutes
    STR_1882    :Every 45 minutes
    STR_1883    :Every hour
    STR_1884    :Every 2 hours
    STR_1885    :Never
    ...
    STR_2701    :Every minute
    STR_2702    :Every 5 minutes
    STR_2703    :Every 15 minutes
    STR_2704    :Every 30 minutes
    STR_2705    :Every hour
    STR_2706    :Never

Overv:
Let me know if you think it's best to remove the non-confirmed ones and wait for native speakers here to handle them.

Gymnasiast:
In that case I'll assume they are correct. Thanks for the explanation.

Gymnasiast:
OK, I merged them. Thanks for the PR. The updated strings will be in the game at 5:00 CEST.

Overv:
Sorry for not being clear enough about it upfront. I realise that "based on" sounds a little unreliable.

Gymnasiast:
Sure, but I have to admit I didn't check what strings 1879 to 1885 were either. Anyway, thanks for your work.

PR #268, opened 2015-10-17
FR: Updated translation

s0r00t:


Gymnasiast:
Thanks, that's a nice update. I found two things:
`-STR_0894    :Landscape data area full !`
"Landscape data area" is a technical structure in RCT2 and OpenRCT2. It is better to keep this untranslated, as it will be far easier to help people getting this error.

`+STR_1171    :{RED}Fermé - -`
This one _needs_ a trailing space, as not doing that will cause the banners to display `Fermé - -Fermé - -Fermé`, etc.

Gymnasiast:
You can translate the 'full' bit, you only need to leave 'Landscape data area' intact.

PR #270, opened 2015-10-19
DE: New strings

danidoedel:


Gymnasiast:
The Condor Ride description is incorrect. It does not use a single rail ('Einschienenstrecke'). This error is present in all original translations, which is exactly why English has an override for it in the first place. Please use the version in the UK English as the translation source.

danidoedel:
Oh sorry missed that... I fixed it now.

PR #272, opened 2015-10-21
NL: Theme names

Broxzier:
Split 'thema' from the theme names


Gymnasiast:
Grammaticaal incorrect.

PR #274, opened 2015-10-27
More improvements to french.txt

s0r00t:


Gymnasiast:
Strings 1993, 2021, 2049 and 2077 refer to fizzy drinks, like cola. I think 'boisson' is a bit too generic a translation.

s0r00t:
Just opened RCT2. Indeed, it's "boissons gazeuses" in French. Fixing it.

PR #277, opened 2015-10-30
Update german.txt

raidcookie:
* Zero clearance text was misleading and read "reset heights"
* Changed tooltip text for Saving/Options as it read "Hard drive and game options"

### DIFF
--- a/data/language/german.txt
+++ b/data/language/german.txt
@@ -834,8 +834,8 @@ STR_0829    :{SMALLFONT}{BLACK}Fenstertitel - Ziehen Sie hier, um das Fenster zu
 STR_0830    :{SMALLFONT}{BLACK}Ansicht heranzoomen
 STR_0831    :{SMALLFONT}{BLACK}Ansicht wegzoomen
 STR_0832    :{SMALLFONT}{BLACK}Ansicht um 90{DEGREE} im Uhrzeigersinn drehen
-STR_0833    :{SMALLFONT}{BLACK}Spielpause
-STR_0834    :{SMALLFONT}{BLACK}Festplatten- und Spieloptionen
+STR_0833    :{SMALLFONT}{BLACK}Spiel pausieren
+STR_0834    :{SMALLFONT}{BLACK}Speicher- und Spieloptionen
 STR_0835    :Spielinitialisierung fehlgeschlagen
 STR_0836    :Spiel kann nicht im minimierten Zustand gestartet werden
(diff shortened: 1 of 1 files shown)


Gymnasiast:
The old text said 'disable height checks', the new one says 'disable building limits'. I don't think this is better, as the old captured the meaning better, IMO. (Yes, I can read German.)

IntelOrca:
* Its not Zero clearance, its disable clearance checks
* Its not related to heights
* It should be the best way of saying, no building / placements checks (no limitations / no constraints, flexible)

raidcookie:
* "The old text said 'disable height checks'"
Yeah, sorry, my mistake. I still think that it'd be misleading since it could be confused with the 'disable support limits' cheat

Gymnasiast:
@IntelOrca : 'Height' and 'clearance' are the same word in German and Dutch. Because of that, `Höhenüberprüfung deaktivieren` can mean either `disable height checks` or `disable clearance checks`/

IntelOrca:
@Gymnasiast perhaps a less ambiguous word could be chosen then?

Gymnasiast:
@raidcookie : Can you think of a German string that boils down to "Don't check if objects are in the way"?

raidcookie:
I still think that the one I chose comes the closest to explaining what disabling clearances does. Besides, clearance rather means "gap" or "play" than height. So 'disable Building restrictions', as I chose, would be the best choice as it is basically what it does.

Gymnasiast:
"Disable building restrictions" is a lot more vague and just as ambiguous with relation to "Disable support limits" as the old one. That's not an improvement.

raidcookie:
It's not about disabling the height restrictions of rides. It's the string for disabling clearances.

Gymnasiast:
I **know**. I'm saying it is just as ambiguous.

IntelOrca:
@Gymnasiast its more vague but less ambiguous and in this context, that is better.

Gymnasiast:
@IntelOrca : I disagree that 'Disable building limits' (the new translation) is less ambiguous than 'Disable height/clearance checks' (the old one).

IntelOrca:
@Gymnasiast ok - we'll have to outsource another opinion then.

danidoedel:
I would change it to `Konstruktionsüberschneidungen erlauben`
That's very clear and accurate, IMO.

Gymnasiast:
@danidoedel : I agree.
@raidcookie : Would you change this?

raidcookie:
Can do.

PR #307, opened 2015-11-22
preface Fahrenheit with degree symbol

e-foley:
Adds a degree symbol before Fahrenheit units to match preferred formatting.


Gymnasiast:
I'm hesistant to merge this, considering you changed _all_ languages, without properly consulting the translators whether it's actually preferred in their respective languages. Considering Fahrenheit is a trivial measure and has no official usage, this can easily vary between languages.

Short research on my part reveals that using a degree symbol seems to be preferred in US English and Dutch, but I would like to hear from the other translators first.

e-foley:
I'm not sure what you mean when you say "Fahrenheit is a trivial measure and has no official usage," but I agree with everything else.  I'll be more careful next time and submit the matter as an issue first.

For the benefit of people like me in the future, can we please add to the readme document some text about consulting with translators before submitting changes?  I can't speak for everybody, but I would rather have proactive guidance than reactive.

Gymnasiast:
What I mean is that there is no official institution behind Fahrenheit, since it's not part of SI. So there is no official use, just common uses, which can very greatly between countries.

As for translator guidelines: that's something I have been meaning to create for some time, but haven't gotten around yet.

As far as the issue is concerned, I'll leave this open for a bit. If no translators complain, I'll merge it for all languages.

e-foley:
Got it, and sounds good. Sorry for the trouble.

IntelOrca:
This can be changed for **en-GB** as well in the main repository.

PR #314, opened 2015-12-02
Created japanese file with some basic translations

nickhall:
Chinese and Korean translations are up, but no Japanese! We can fix this.

### DIFF
--- a/data/language/japanese.txt
+++ b/data/language/japanese.txt
@@ -1,0 +1,3931 @@
+# STR_XXXX part is read and XXXX becomes the string id number.
+# Everything after the colon and before the new line will be saved as the string.
+# Use # at the beginning of a line to leave a comment.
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Thanks for your contribution!
Could you suggest a Japanese font to use that is likely to be installed on most Windows computers?

nickhall:
Meiryo or MS Gothic might be good choices!

AaronVanGeffen:
MS Gothic is the most compatible choice, I think -- it's available from XP onwards. Meiryo has been available since Vista, but is mostly for use with anti-aliasing / cleartyping, I think -- which OpenRCT2 currently does not support.

IntelOrca:
It would be nice to use DirectWrite in the future so we can properly do ClearType text.

PR #315, opened 2015-12-04
more translation, strings revision and more

renansimoes:
more translation, strings revision and added some missing lines


renansimoes:
could you guys please add a new Currency?

Real Brasileiro (R$)

PR #317, opened 2015-12-08
translation ptbr complete and string revision

renansimoes:
all new strings were added and string issues revised


Gymnasiast:
I merge this one manually, but you still need to change these two:

PR #322, opened 2015-12-13
ptbr strings 5525/5526 changed

renansimoes:
As you said, I changed the strings 5525 to "Água escura" and 5526 to "Água clara"


Gymnasiast:
Thanks!

PR #338, opened 2015-12-27
added strings

renansimoes:
added and translated strings: 5581, 5582, 5583, 5584, 5585, 5588
new empty strings: 5586, 5587


Gymnasiast:
I won't merge the file with 5586 and 5587 being empty, as that will cause empty places in the interface. Please either translate them or remove them completely, so they show the English strings instead.

renansimoes:
Why not? These strings (5586 & 5587) still doesn't exist at English file,
5588 hasn't been added to that file either. 5588 is at issues topics.
Em 27/12/2015 17:07, "Michael Steenbeek" <notifications@github.com>
escreveu:

Gymnasiast:
That's simply not true. Unless you're looking at english_us.txt, but that is **not** the file you should base anything on. The correct file to look at is english_uk.txt, which has those strings.

renansimoes:
Sorry, I really was looking at english us file x.x

Gymnasiast:
Good, thank you. :)

PR #340, opened 2015-12-28
Some Spanish Translations

teapartycthulu:
About 783-825: I think these are the ones used for dates, so I left them
in proper form. It should be "el {daynumber} de {month}" when formatted.
(either abbreviated or not)
I'm not very good at my Spanish, so this is just a first run through of
things I know for sure. More to come soon.
Also this might be more Latin/Mexican Spanish than Spanish Spanish, as the file name suggests. Might need a change there later.


Gymnasiast:
Considering nobody else has stepped up, having some Mexican Spanish or some slight errors in the European Spanish translation is the least of my worries. Thanks for picking it up. :+1:

PR #351, opened 2016-01-04
Update korean.txt

xiane:
- Make korean translation more naturally.
- changed translation of "checksum"

- Here is suggestion.
- translate "checksum" to "채크섬(썸)" or "검사 합".
- How about this? please check and answer about it.

- Please do not merge it before end of discussion.


telk5093:
'검사합'으로 번역하겠습니다. 그리고 깃헙에서 그냥 직접 번역하시면 됩니다.
I'll translate checksum into '검사합' and you can translate it in github.

PR #356, opened 2016-01-08
Translated some strings to Japanese

jhako:

PR #361, opened 2016-01-13
JP: A bunch of miscellaneous additions

nickhall:
I went through at random and added a bunch of things that I felt I could translate with a reasonably high degree of confidence (since I'm not a native speaker). I'll try to consult with native speakers to translate some of the really tricky stuff if nobody else gets to it in the mean time.


AaronVanGeffen:
Looks good to me.

jhako:
I'm native, and it seems natural :)

AaronVanGeffen:
@jhako Thank you. Incidentally, if you have time to translate strings, your help would be most welcome! :smiley:

nickhall:
Apparently I wrote 来客の沖にいる though.. I'll fix that too

jhako:
@AaronVanGeffen Yes, I'll join the effort. :smile:

@nickhall 
I think 来客のお気に入り is good instead of it
and "停車時間" would be better than "ダウンタイム"

nickhall:
@jhako Downtime was the only one I was really unsure about, but it was [the first result on alc](http://eow.alc.co.jp/search?q=downtime&ref=sa) so I went with it anyway. I'll submit a pull request to fix the お気に入り one and change ダウンタイム. I'm really curious to see how you would translate some of those coaster names.

Gymnasiast:
Could you include the changes in this PR, where appropriate: https://github.com/OpenRCT2/Localisation/pull/356

nickhall:
@Gymnasiast Is there a way to merge that into this? I'm no git pro but I can't seem to find a way to get his commits into my branch.

Gymnasiast:
You can just edit the file in your patch-1 branch and copy and paste the changed strings in that commit.

nickhall:
I guess so. Any reason you don't want to accept both PRs? That way it would preserve his contributions at least.

Gymnasiast:
That required rebasing, which isn't really easy in this case. But I did it anyway to keep his name on his commit.

PR #364, opened 2016-01-14
Translated some strings to Japanese

Gymnasiast:


jhako:
That seems hard work!
Thank you.

PR #365, opened 2016-01-16
KOR : #357,#359,#360

NeverDruid:
added strings.

PR #366, opened 2016-01-16
KR: #357, #359, #360 and more translations

telk5093:
Applied #357, #359, #360 as @NeverDruid did in #365.
And translated more strings and fixed some typos.

### DIFF
--- a/data/language/korean.txt
+++ b/data/language/korean.txt
@@ -21,7 +21,7 @@ STR_0016    :전망탑
 STR_0017    :루핑 롤러코스터
 STR_0018    :미니 슬라이드
 STR_0019    :마인 트레인 코스터
-STR_0020    :체어리프트
+STR_0020    :체어 리프트
 STR_0021    :코르크스크류 롤러코스터
 STR_0022    :미로
 STR_0023    :나선 미끄럼틀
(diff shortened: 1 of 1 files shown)


data/language/korean.txt
```diff
@@ -11255,7 +11257,7 @@ STR_SCNR    :유럽 - 혁신
 STR_PARK    :잿더미를 딛고
 STR_DTLS    :방치되어 있는 낡은 공원이 있습니다. 이 공원이 과거에 누렸던 영광을 재현하기 위해 유럽 연합의 허가를 받았습니다! 허가를 받은 만큼 공원을 혁신해서 보답하세요.
 
-<N America - Extreme Hawaiian Island>
+<N. America - Extreme Hawaiian Island>
```
Gymnasiast: Change this back. This is a filename, which you shouldn't change.

telk5093: But it does not be translated if this dot is missed.

I attached that the image using this korean.txt language file
image

Gymnasiast: Hm, that's strange. In that case you can leave it as it is.

telk5093: Isn't it the scenario's name, not the file name?

Gymnasiast: Probably. It must have been changed with the introduction of the new scenario dialog.

PR #370, opened 2016-01-22
JP: More progress

nickhall:
I found 暴走トロッコ on alc and it seems apt, so I went ahead and threw it in.

### DIFF
--- a/data/language/japanese.txt
+++ b/data/language/japanese.txt
@@ -977,7 +977,7 @@ STR_0971    :+495{DEGREE}
 STR_0972    :キャンセル
 STR_0973    :OK
 STR_0974    :乗り物
-STR_0975    :お店と屋台
+STR_0975    :Shops and Stalls
 STR_0976    :トイレと案内所
 STR_0977    :New Transport Rides
 STR_0978    :New Gentle Rides
--- a/data/language/japanese.txt
+++ b/data/language/japanese.txt
@@ -16,11 +16,11 @@ STR_0011    :Wooden Wild Mouse
 STR_0012    :Steeplechase
 STR_0013    :Car Ride
 STR_0014    :Launched Freefall
-STR_0015    :Bobsleigh Coaster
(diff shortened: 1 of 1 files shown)


data/language/japanese.txt
```diff
@@ -976,9 +976,9 @@ STR_0970    :+450{DEGREE}
 STR_0971    :+495{DEGREE}
 STR_0972    :キャンセル
 STR_0973    :OK
-STR_0974    :Rides
-STR_0975    :Shops and Stalls
-STR_0976    :Toilets and Information Kiosks
+STR_0974    :乗り物
+STR_0975    :お店と屋台
```
AaronVanGeffen: First off, wouldn't 乗り物 be a vehicle rather than the actual ride?

Possibly nitpicking, but at least for consistency: earlier you transliterated "Shops" using katakana (ショップ). Whichever you choose, I'd rather use the same throughout. I'm also not convinced お店 carries the same meaning?

nickhall: Actually I was debating a lot about which words to use for some of these and ended up changing most of them back, but I missed one apparently. For example for both food shop and drink shop I'd say 屋台, but then for things like souvenirs and whatnot...? We need to hammer out those details.

乗り物 is acceptable for amusement park rides.

AaronVanGeffen: 屋台 sounds good!

Gymnasiast:
Is this ready for merging?

nickhall:
I switched back one of the more suspect translations. Those specifics can be addressed later. Should be good for a merge now.

nickhall:
Sneaked in a few roller coaster translations, but these shouldn't be too controversial. Taken from https://ja.wikipedia.org/wiki/%E3%83%AD%E3%83%BC%E3%83%A9%E3%83%BC%E3%82%B3%E3%83%BC%E3%82%B9%E3%82%BF%E3%83%BC#.E7.A8.AE.E9.A1.9E

PR #380, opened 2016-01-26
CZE: (last) batch of strings translated

octaroot:
This should be the last batch of strings being translated to Czech. Right now, the only strings that are left untranslated are the names+descriptions of some of the rides/attractions - which are a real pain to translate/I'm not able to translate them accurately. Otherwise - after 3 days, everything else is now in Czech :-)

With 2k+ strings translated, it would be nice to be included in contributors file, it at all possible (don't know whether some special rules apply)

PR #383, opened 2016-01-26
CZE: .DAT strings added & partially translated

octaroot:
I added all of .DAT strings to the czech.txt file (since RCT2 was never *officially* translated to Czech) and I translated the first ~2.5k section titled `RCT2 Objects` - that amounts to about a third of .DAT strings.

Oh boy oh boy, *so* many strings!

Sorry about *yet another* pull-request, but as @Killermannvs mentioned in pull request #380, a certain degree of cooperation between translators is required, when working in parallel.

### DIFF
--- a/data/language/czech.txt
+++ b/data/language/czech.txt
@@ -4080,123 +4080,123 @@ STR_CPTY    :10 míst v každém vozu
 
 [4X4]
 STR_NAME    :Monster Trucks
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Thanks! And don't worry about opening too many PRs: we're very happy that you translate OpenRCT2.

PR #388, opened 2016-01-29
CZE: .DAT Scenarios (RCT2 & WW & TT) strings translated

octaroot:
And here it is - the final batch of translations - all game scenarios. The game and all of it's content should now be translated into Czech. I'm sure I've overlooked some stuff, so some testing (by the Czech community?) will probably be required before calling this "fully translated", but the majority of work is now done. It took many, many hours. I'm so glad it's over :D


octaroot:
Oh, I see this PR also includes commits from PR #387 ... will this be a problem?

Gymnasiast:
It wasn't a problem because I didn't merge #387.

Thanks for your work!

PR #412, opened 2016-02-11
Update spanish_sp.txt

Wirlie:


Wirlie:
@gdabi you can take a look for these commits?

Wirlie:
@Gymnasiast please merge it. I translated these strings using original translations of RCT2-Spanish. I tested these translations in-game.

But.... @gdabi has not answered my emails from 5 days ago :/

I will contact gdabi when he come back.

data/language/spanish_sp.txt
```diff
@@ -4047,114 +4050,114 @@ STR_5736    :Jugador
 
 ## RCT Original
 <Forest Frontiers>
-STR_SCNR    :Forest Frontiers
-STR_PARK    :Forest Frontiers
-STR_DTLS    :Deep in the forest, build a thriving theme park in a large cleared area
+STR_SCNR    :Fronteras Forestales
```
Gymnasiast: Were the scenario names translated in the original Spanish RCT1?

data/language/spanish_sp.txt
```diff
@@ -4483,11 +4486,9 @@ STR_PARK    :Fort Anachronism
 STR_DTLS    :
 
 #########
-# Rides #
+# STALL #
 #########
 
-#WW
```
Gymnasiast: Why did you take this entry out? Please add it back in.

data/language/spanish_sp.txt
```diff
@@ -4483,11 +4486,9 @@ STR_PARK    :Fort Anachronism
...
-[CONDORRD]
-STR_NAME    :Condor Ride
-STR_DESC    :Riding in special harnesses below the track, riders experience the feeling of flight as they swoop through the air in Condor-shaped trains
-STR_CPTY    :4 passengers per car
+[INFOK]
```
Gymnasiast: Is the original RCT2 description of the Information kiosk incorrect in Spanish? You only need to specify an override if the original is badly translated or plainly wrong.

gdabi:
Sorry guys.

It's OK to use the original translations, but I don't have the original RCT2, so feel free to override my translated strings.

Gymnasiast:
Original strings aren't necessarily better. Since I know hardly anything about Spanish, I assume that you (Wirlie and gdabi) make that judgement, rather than blindly replacing everything with the original translations.

Anyway, I left some remarks on the PR for Wirlie to look at.

Gymnasiast:
Original strings aren't necessarily better. Since I know hardly anything about Spanish, I assume that you (Wirlie and gdabi) make that judgement, rather than blindly replacing everything with the original translations.

Anyway, I left some remarks on the PR for Wirlie to look at.

gdabi:
@Gymnasiast that's true. I remember when I was young and was playing RCT2 thinking that some strings doesn't fit exactly what they want to say, and there is another problem too, that there is some "more technical" strings (like Block brakes) which cannot be translated literally.

Anyway, the original translation is very good (for the spanish case), so we can use it as a start point. Later, when a more flexible toolkit is implemented, will make us possible to generate better translations.

Gymnasiast:
@gdabi : I vastly improved the Dutch translation in this 'toolkit', so it seems flexible enough to me for this purpose. (Although I admit there are better solutions.) The original translations for Dutch, German and some (most?) CJK translations were very bad. Good to see that is not the case for Spanish.
Anyway, I trust you two to know what's best for the Spanish translation.

data/language/spanish_sp.txt
Wirlie: original translation show "Inf**r**omacion" not "Info**r**mación"

data/language/spanish_sp.txt
Wirlie: Yup!

data/language/spanish_sp.txt
Wirlie: Reverted in ba406c4, thanks!

Wirlie:
@Gymnasiast Thanks! I have made the necessary changes. Please take a look.

data/language/spanish_sp.txt
```diff
@@ -4482,12 +4487,11 @@ STR_SCNR    :Fort Anachronism
...
 
-#########
-# Rides #
-#########
+#####################
+# Rides/attractions #
+#####################
 
 #WW
-[CONDORRD]
```
Gymnasiast: This block is still missing. Please re-add it. (The original English text was wrong, as it described the wrong type of track, therefore the Spanish translation is wrong too.)

Wirlie: Oh! I understand, thanks! I added again this section.

Wirlie:
Okay, thanks very much for your patience. Now I know how to work with the next updates :)

PR #443, opened 2016-02-25
DE: new strings

danidoedel:
I'll change `STR_5766    :{SMALLFONT}{BLACK}Paint land height` later when I have more time to find it ingame...


IntelOrca:
Its not used, just change it to ```<available string id>```.

danidoedel:
done

PR #474, opened 2016-03-04
Update swedish + new strings 2016/03/14.txt

Nubbiy:
*Fixed spelling errors
*Filling in missed strings

*Added the new strings 
*Added myself as a maintainer in this document 
*Translated some new strings (50 %)
(*Not to be a douche - but corrected the spelling of "Maintainer" on the Note)

Overall progress: (75 %)

QUESTION: What is the difference between string 5649 & 5651?

### DIFF
--- a/data/language/swedish.txt
+++ b/data/language/swedish.txt
@@ -64,7 +64,7 @@ STR_0059    :Inverterad berg- och dalbana
 STR_0060    :Okänd Åktur (3A)
 STR_0061    :Virginiavirveln
 STR_0062    :Plaskbåtar
-STR_0063    :Mini Helicopters
+STR_0063    :Mini Helikoptrar
 STR_0064    :Liggande berg- och dalbana
 STR_0065    :Hängande balkbana
 STR_0066    :Okänd Åktur (40)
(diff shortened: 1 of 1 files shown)


Nubbiy:
I'm really thankful for this mod and the work you guys are putting down so I want to help out too, sadly enough I don't know how to script but I would be more than happy to translate and maintain the Swedish language on OpenRTC2, it's not much but something.

Gymnasiast:
Thank you for your contribution. :)

The difference between strings 5649 and 5651 is as follows:
5649 is about creating a new ride entity.
5651 is about constructing a ride, whether that is an old or a new ride.

You can get rid of the maintainer notes altogether, as the language is now maintained. The note is there because unmaintained language files are minimised, which means that untranslated strings are removed to make them fall back on English, making less likely to break when strings are changed.

Nubbiy:
Thank you :)

Oh I see, will translate that immediately

I would, and I would be more than happy to continue translating this file, the problem is that I have no clue how as I don't use this site too often :(

Gymnasiast:
If you look at the top if the PR you can see your edits are in your patch-2 branch. So you go to your own Localisation repository at https://github.com/Nubbiy/Localisation , then you click the branch dropdown and select patch-2. Then you navigate to the file you want to edit, and make your edits. As soon as you save that file, this PR will be updated.

Gymnasiast:
I'm merging this now, as a stable is coming up. If you have any improvements the following days, we can still have them in stable btw.

Nubbiy:
Thanks for telling me how to edit my branches, but have been a bit busy so couldn't modify the file, will try and translate more today

Issue #476, opened 2016-03-04
STR_5789


danidoedel:
OpenRCT2/OpenRCT2@57052d6

added string:

`STR_5789    :Disable lightning effect`

PR #478, opened 2016-03-05
KOR : Apply #476

telk5093:
Added some alias of scenarios becuase it doesn't work correctly.
I will remove when this issue fixed


Gymnasiast:
`Haunted Harbour` and `Mythological - Cradle of Civilisation` are the correct names as far as OpenRCT2 is concerned. Please change those entries in your file.

Also, if you have a file called Utopia.SC6, that file is incorrectly named.

IntelOrca:
This was mentioned in https://github.com/OpenRCT2/OpenRCT2/issues/3093

telk5093:
I removed them.
And http://www.mediafire.com/download/3cw47sd2wt3hues/ExactRCT1Recreations.zip in pull(#2628) has Utopia.SC6. I think that was wrong.

Issue #484, opened 2016-03-15
STR_2775 and STR_5177


Gymnasiast:
```
STR_2775    :Fullscreen (borderless window)
STR_5177    :Screen mode:
```

IntelOrca:
US still needs updating

dlee13:
I think a diff should be run on UK and US. Then remove the strings that are identical, and leave the strings that are different. That way, when UK is changed, US has its strings changed as well. I don't see the reason for having the duplicates in US when it can fallback on UK.

Gymnasiast:
Indeed. Although I would have to check how it behaves with low IDs.

Gymnasiast:
I have removed strings in US with IDs higher than or equal to 5177, except for those that differ from UK English. It will therefore now show the correct strings.

PR #490, opened 2016-03-19
CHT: #484

daihakken:
5177 is now "Display Setting" in Chinese for clearer meaning to Chinese reader (Screen Mode doesn't quite make sense in Chinese).


Gymnasiast:
That's a good decision, as the meaning of the string is most important.

PR #493, opened 2016-03-25
FR date translation update

rvgr:
Dates translation is now in the correct French format.

Was: 24ème Sept, an 1
Now: 24 Septembre, année 1


Gymnasiast:
I seem to remember months not being capitalised in French:
`24 septembre, année 1`
Are you sure this is correct?

rvgr:
You're right, I overlooked it. Corrected!

PR #499, opened 2016-03-30
Updated Swedish.txt - 95-100% translated (2016-03-30)

Nubbiy:
Every single string is translated, some small changes are made to fit inside the layout of the game, spelling corrections

Todo; Check in the newest version for errors and correct spelling mistakes if found

### DIFF
--- a/data/language/swedish.txt
+++ b/data/language/swedish.txt
@@ -2706,7 +2706,7 @@ STR_2696    :Placera trä
 STR_2697    :???
 STR_2698    :???
 STR_2699    :???
-STR_2700    :Autospar frekvens:
+STR_2700    :Spara automatiskt:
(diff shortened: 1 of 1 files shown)


data/language/swedish.txt
```diff
@@ -4547,7 +4547,7 @@ STR_DTLS    :
 #WW
 [CONDORRD]
 STR_NAME    :Condor Ride
```
Gymnasiast: Are you sure you don't want to translate the ride name?

Nubbiy: I saw that right now, dang it, missed it :p 
Will have to correct that later on when I test the text out to see if it fits inside the boarders and when I search for spelling mistakes etc
Don't want to spam you with pull requests

PR #482, opened 2016-03-10
Updated japanese.txt

RapleAQN:
Changed some strings can't do more at the moment

PR #518, opened 2016-04-07
Japanese translation update

AaronVanGeffen:
This resolves the discussion in #482, giving credit to the original translators.

### DIFF
--- a/data/language/japanese.txt
+++ b/data/language/japanese.txt
@@ -76,12 +76,12 @@ STR_0071    :ロトドロップ
 STR_0072    :空飛ぶ円盤
 STR_0073    :ねじれた家
 STR_0074    :Monorail Cycles
-STR_0075    :Compact Inverted Coaster
-STR_0076    :Water Coaster
-STR_0077    :Air Powered Vertical Coaster
-STR_0078    :Inverted Hairpin Coaster
+STR_0075    :コンパクト倒立コースター
+STR_0076    :ウォーターコースター
+STR_0077    :エアパワード垂直コースター
+STR_0078    :倒立ヘアピンコースター
 STR_0079    :魔法のじゅうたん
-STR_0080    :Submarine Ride
+STR_0080    :潜水艦ライド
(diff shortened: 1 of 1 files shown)


nickhall:
I may have missed some of the remaining katakana transliterations in the previous thread, but I have to question some of them, especially something like ジェントルライド. Do we want to use ライド and not 乗り物? I'd suggest 穏やかな乗り物. Similarly, instead of エアパワード maybe 空気動力? For 潜水艦ライド maybe delete the ライド and use only 潜水艦? トランスポート could be 輸送 or 交通. Some of these may be passable with their katakana equivalents, including ones I suggested, but I'd be interested in seeing what others think.

jhako:
I think it's one of the good ways for translation to avoid using katakana as possible, but it can be unnatural for Japanese. Katanaka is, in my opinion, often used in words of things which is modern (about last 50 years?) and came from foreign, and also usually used in proper names. I feel 穏やかな乗り物 like common names of calm vehicles. So I think it's better to use not 乗り物 but ライド ,for 乗り物 is more general words. For proper names, I'd suggest 穏やかライド, 優しさライド or 安心ライド. Both of 潜水艦ライド and 潜水艦 are good, first one may be repetitive and second one may be too proper (army?), a matter of preference. トランスポート would be 移動手段 or 移動用ライド. エアパワード could be 空気力駆動 but this might be like technical terms so that I like katakana one or 空気パワー.

nickhall:
@jhako  I'm surprised to hear that you'd prefer ライド. I didn't really see it in my searches, but it's good to know. I felt the same about overuse of kanji. Something like 空気パワー is a great balance. For transport I was going crazy trying to think of the right word for it, and 移動 was what I was trying to connect to it. Thanks as always for your contributions.

Gymnasiast:
A release is imminent. I hope you can reach consensus by tomorrow (Western European time).

Gymnasiast:
I have decided to merge it as it is because of the release, but please continue with the discussion.

PR #525, opened 2016-04-10
DE: New string

danidoedel:


Gymnasiast:
Your translation suggests scenery groups rather than a cluster.
The string refers to whether the user is allowed to use this button:

scr185

Issue #535, opened 2016-04-17
STR_5759 has units now


janisozaur:
https://github.com/OpenRCT2/OpenRCT2/issues/3328

`KiB`

PR #536, opened 2016-04-17
SE: 100% translated + Added str from #535

Nubbie:
Translated it all, proof-read it and spell checked it with Libre office proof-reading software
Added the string from #535


Nubbie:
Guess I dont have to worry about the check that failed

Gymnasiast:
Spell check frequently marks words that are spelt correctly as wrong. Especially compound words, uncommon words and grammar. It's usually better to let a human check it. Anyway, as I have little knowledge of Swedish, I'll just assume it's correct.

Nubbie:
@Gymnasiast Indeed which I'm totally aware of, which is why I manually check and changed stuff twice, didn't let the program go through everything itself as that would be horrible :p
Was a lot that the program had wrong

PR #543, opened 2016-04-19
CHS: update setting panel translation

izhangfei:


izhangfei:
why my pull request have lots of commits?

Gymnasiast:
I'm not sure why. I squashed them into one when committing.

Issue #548, opened 2016-04-19
STR_5755 wants a string now


janisozaur:
This is to hint server version to the client, when they are incompatible.

https://github.com/OpenRCT2/OpenRCT2/pull/3350

```
STR_5755    :Incorrect Software Version (Server is using {STRING})
```

Wirlie:
For more clarification, this is the field with this change:

    STR_5755    :Incorrect Software Version (Server is using {STRING})

Issue #545, opened 2016-04-19
STR_5801 and STR_5802


Nubbie:
Added strings from [#3152](https://github.com/OpenRCT2/OpenRCT2/pull/3152)
```
STR_5801    :Disable littering
STR_5802    :{SMALLFONT}{BLACK}Stops guests from littering and vomiting
```

PR #549, opened 2016-04-19
CZE: Apply #545 #548

octaroot:


Nubbie:
@octaroot Just want to tell you that #551 adds:

`STR_5803    :{SMALLFONT}{BLACK}Rotate selected map element`

PR #566, opened 2016-04-24
New strings translated

LucaRed:
I've translated some strings and fixed some typos.


Gymnasiast:
Thanks!

LucaRed:
I've seen at the end of the document that Italian is not maintained, so we shouldn't add new strings.

I'd like to maintain the Italian language, what should I do? (Sorry if this is not the place to ask, I'm a total newbie on GitHub)

Gymnasiast:
Thank you! In that case you can remove the string. :)

PR #581, opened 2016-04-30
it-IT: New strings translated

LucaRed:


Nubbie:
See you added some new strings and that is why I just want to tell you 3 more strings have been added, see issue #583 :)

LucaRed:
Thanks! I've added the new strings :smile::+1:

PR #610, opened 2016-05-06
few italian translations

Mte90:
I hope to continue with the translations in the next days :-)


Nubbie:
I hope you keep on translating the language :)

Would be wonderful if you could add + translate these strings too that are missing from the document

```
STR_5806    :Toggle windowed mode

STR_5810    :Disable train length limit
STR_5811    :{SMALLFONT}{BLACK}If checked, you can have up to{NEWLINE}255 cars per train
```
Just remember to put them in the right STR order :p

Mte90:
Okay!
I will work on that translation in the spare time :-)

Issue #611, opened 2016-05-06
STR_5812


Nubbie:
Added string from: [#3509](https://github.com/OpenRCT2/OpenRCT2/pull/3509)
```
STR_5812    :Show multiplayer window
```

Nubbie:
Just two (maintained) languages missing this string now:
it-IT.txt
pt-BR.txt

Mte90:
Italian translation done :+1:

PR #612, opened 2016-05-06
Apply #611

Nubbie:
#611


Nubbie:
um.. ignore all these things above this

Issue #622, opened 2016-05-08
STR_3357, STR_5813 and STR_5814


danidoedel:
https://github.com/OpenRCT2/OpenRCT2/commit/fc6a589c3551917da5b5ab744d723c690011338f
https://github.com/OpenRCT2/OpenRCT2/pull/3477

**changed string:**
```
STR_3357    :{WINDOW_COLOUR_2}Are you sure you want to permanently delete {STRING} ?
```

**added strings:**
```
STR_5813    :{OPENQUOTES}{STRING}{ENDQUOTES}
STR_5814    :{WINDOW_COLOUR_1}{OPENQUOTES}{STRING}{ENDQUOTES}
```

Nubbie:
Double checked, and all the languages now have `STR_3357` changed, all maintained languages have `STR_5813 + STR_5814` added to it
This issue maybe should be closed to not confuse people :p

PR #626, opened 2016-05-08
Apply#622 for all languages

Nubbie:
**CLOSE THIS PR IF YOU THINK THIS WAS WRONG OF ME**
#622


Gymnasiast:
No, you did good work.
But could you remove STR_5813 and STR_5814 (only those two, changing STR_3357 is good) from the following unmaintained or badly maintained languages:

ja-JP
ru-RU
hu-HU
fi-FI
pl-PL

Nubbie:
Perfect, and here I was thinking I would get a scolding :p

Issue #625, opened 2016-05-08
STR_5815 -> STR_5859


Nubbie:
Added tool-tips strings from [#3394](https://github.com/OpenRCT2/OpenRCT2/pull/3394) _and with correction of [#3534](https://github.com/OpenRCT2/OpenRCT2/pull/3534)_

**New added strings:**

```
#tooltips
STR_5815    :{SMALLFONT}{BLACK}Display FPS counter in-game
STR_5816    :{SMALLFONT}{BLACK}Sets ratio to scale the game by.{NEWLINE}Most useful when playing in{NEWLINE}high resolutions
STR_5817    :{SMALLFONT}{BLACK}[Requires hardware display]{NEWLINE}Sets UI scaling type
STR_5818    :{SMALLFONT}{BLACK}[Requires hardware display]{NEWLINE}Use nearest neighbour scaling{NEWLINE}when window scaling factor set{NEWLINE}to integer values (1, 2, 3, etc)
STR_5819    :{SMALLFONT}{BLACK}[Requires hardware display]{NEWLINE}Pause the game if Steam{NEWLINE}in-game overlay is opened
STR_5820    :{SMALLFONT}{BLACK}Minimise the game if focus is{NEWLINE}lost while in fullscreen mode
STR_5821    :{SMALLFONT}{BLACK}Changes the colour of the construction marker when building rides,{NEWLINE}paths, shops, scenery, etc.
STR_5822    :{SMALLFONT}{BLACK}Cycle between day and night.{NEWLINE}Full cycle takes one in-game month
STR_5823    :{SMALLFONT}{BLACK}Display banners in uppercase (RCT1 behaviour)
STR_5824    :{SMALLFONT}{BLACK}Disables lightning effect{NEWLINE}during a thunderstorm
STR_5825    :{SMALLFONT}{BLACK}Keep the mouse cursor in the window
STR_5826    :{SMALLFONT}{BLACK}Invert right mouse dragging of the viewport
STR_5827    :{SMALLFONT}{BLACK}Sets the color scheme used for the GUI
STR_5828    :{SMALLFONT}{BLACK}Changes what measurement format is used for distances, speed, etc.
STR_5829    :{SMALLFONT}{BLACK}Changes what currency format is used. Purely visual, there is no exact exchange rate implementation
STR_5830    :{SMALLFONT}{BLACK}Changes what language is used
[...]

PR #627, opened 2016-05-08
cs-CZ: Apply #625

octaroot:
Removing STR_5813/5814, not sure what those are doing there, since they are covered by #625.


Nubbie:
YOU just cracked the code why it doesn't work as it should! Thanks!

octaroot:
OK :) Done

PR #638, opened 2016-05-13
JP: Miscellaneous additions

nickhall:
I tried to break it apart into a few different commits. As always feel free to disagree with any choices. Again based mostly on the RCT3 translation wiki.


nickhall:
One possible point of discussion: use 研究する or 開発する? @jhako

Gymnasiast:
@AaronVanGeffen

jhako:
研究する is better. 開発する means I think we can use new rides at once.
Great works :)

nickhall:
@Gymnasiast User @jhako is a native speaker and gave his endorsement of this PR so I'd say it's safe to merge.

Issue #645, opened 2016-05-18
STR_3160


octaroot:
Changed string from OpenRCT2/OpenRCT2#3650.
```STR_3160    :{SMALLFONT}{BLACK}Select the number of circuits per ride```

Updated in:
- [x] en-GB
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] sv-SE
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW

The following unmaintained languages need the string removed (or translated) to have it fall back:
- [x] pl-PL
- [x] ja-JP
- [x] hu-HU
- [x] fi-FI
- [x] ru-RU
- [x] it-IT

Wirlie:
done, you can mark **es-ES** as updated :P

Nubbie:
en-US is ment for words that differ from UK, so you could remove that
Smart to use a list to keep track of the languages ^^
You can add sv-SE, going to do it right now

octaroot:
@Nubbie True, but this is a changed string, in en-US it's marked as "<not used anymore>" and would show like that in the game, so it needs to be either changed or removed.

Thinking about it, if possible, we should probably remove <not used anymore> from all translations (excluding en-GB), so that it would fall back to en-GB - in cases like this one.
![scrn](https://cloud.githubusercontent.com/assets/16853773/15362492/9635fc54-1d15-11e6-9f64-567001495502.png)

Gymnasiast:
Please, **never** reuse an old string ID again, and certainly not a number under 5120. It prevents automatic fallback, which especially breaks languages that aren't frequently updated.

octaroot:
Oh, OK. The reason I picked that old ID was to get it semi-close to related strings. Won't happen again :-)

Nubbie:
@Gymnasiast We could easily change the strings back in all files and move the string ID to the newest to correct this if you want :) 

@octaroot Yeah that's true as I didn't think about it

Nubbie:
Add it-IT.txt to the list btw

PR #657, opened 2016-05-21
Changing STR_3160 for:   pl-PL + ja-JP + fi-FI + ru-RU (Applying #645)

Nubbie:
As requested in: #645 

**Q&A**
**Q**uestion: Why did I add the strings and not just remove them?
**A**nswer: As these strings will need to be translated anyways later on when someone translates the file, it's better to fill in the blank instead of leaving a empty space easily for someone to miss :U

### DIFF
--- a/data/language/fi-FI.txt
+++ b/data/language/fi-FI.txt
@@ -3171,7 +3171,7 @@ STR_3156    :
 STR_3157    :kartta
 STR_3158    :graph
 STR_3159    :lista
-STR_3160    :<not used anymore>
+
 STR_3161    :<not used anymore>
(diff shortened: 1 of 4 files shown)


Gymnasiast:
Please remove them instead. If the source string gets changed, it might break these languages.

Nubbie:
@Gymnasiast Like I said in the Q&A, I'm worried they will be missed by translators as they are in the middle of the whole pile :U 

Could go for Plan B: Close this and when every language have been adding this string: Move them all the way down for all languages, change the string ID etc

Gymnasiast:
If all untranslated strings are removed, it should be more obvious. I prefer that above copying.

PR #658, opened 2016-05-21
DELETING ru-RU translation (do not merge right away!) (Read alternative too) (Close if a 'No')

Nubbie:
**This is a bigger step and a bigger deal than just adding some strings.**
_Will try and make this short while still making some sense._

I would like to **remove the whole ru-RU translation** from Localisation and Openrct2/data

**Why:**
The translated file only has 130 strings translated (hand counted) out of around 3000 strings (only original strings from the game). The last edit off the file was back in Sep 11, 2015 by a user who seems a bit inactive (only fork is OpenRCT localisation with 3 edits)

For someone who speaks Russian, just having 95% of the coaster names translated + 25 % of their descriptions would be useless when you consider that 130 / 3000 strings are in English.

I presume it's 130 strings out of roughly 3600 translated..

Thoughts? :o 

**My strongest point why:**
Unnecessary for a Russian speaker to only read the ride names and not have anything else translated

Read:
**ALTERNATIVE**: Remove all strings that ain't translated

_(We could still have the original contributor left in the contribution file)_


Gymnasiast:
I prefer the alternative of only leaving untranslated strings. See my recent commit for Hungarian. Please also copy the notice I put on that file.

PR #685, opened 2016-05-30
fi-FI: Apply #645

Nubbie:
It's just one string in the masses, as Finnish is almost 100% translated (but missing new strings), I don't think it would be fair to remove it as it seems like it's something that could easily be missed then.

#645


Nubbie:
Note: Remember what you said about: 
>Please remove them instead. If the source string gets changed, it might break these languages.

Will remove it instead

PR #692, opened 2016-05-31
Create PULL_REQUEST_TEMPLATE

Nubbie:
It's good if you want to keep better track which issues are applied for each language etc
It's a bit off a rough example

This is how it would be easier if they **filled in the issue number in the description:**

merged

**Instead off** 
nono

Or even worse.. **nothing at all**

PR #695 is testing this out right now (My se-SV pr )

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,0 +1,3 @@
+Applying for issue: #0000
+
+Extra info:
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,4 +1,3 @@
-_If you're applying strings from a issue, please fill it out below_
-- Applying for issue: #0000
+If you're applying strings from a issue, please fill it out below (if not, delete this text)
 
-_Additional information/other:_
+- Applying for issue: #0000
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,4 +1,4 @@
-_If you're applying strings from a issue, please fill it out bellow_
+_If you're applying strings from a issue, please fill it out below_
 - Applying for issue: #0000
 
 _Additional information/other:_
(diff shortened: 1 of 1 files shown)


Wirlie:
And what is the problem of specify the number in the commit? ... I prefer to use the number in the commit instead of PR.

Nubbie:
this

Think it gets a little more messy xox

Wirlie:
Okay, then I will not use the template.

Nubbie:
As we now have a list of languages that apply strings, having a larger text that ref the issue + have the text 'Merged' afterward (see screenshot) would just be a bit easier to see xox

Wirlie:
But you are the only does that. xP 
It's a matter of preference.

Wirlie:
Well, I could use the template but I will continue to specifying the number in the commit (as usual). Does that is better?

I don't want to cause a conflict around the world because of a template. >-<
You always try to find the best solution to something. :P

Nubbie:
The template isn't even merged, still a suggestion and all, still not something you have to use :p 

Just want to make it as simple as possible.. ^^ 

I mean jesh, If I was able to program, we would have this:

Wirlie:
Hum, you could add these bars manually to the README... the problem is that there is no way to update it automatically (I think).

bar
bar
bar

(https://github.com/fehmicansaglam/progressed.io)

.github/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,4 @@
+_If you're applying strings from a issue, please fill it out bellow_
```
Gymnasiast: below*

Nubbie:
Would be good to have a standard but have no idea what the text should say :U

Nubbie:
@Gymnasiast 
See some people are using it and some don't, depends if you want to make it a bit easier to link to issues to check them, I presume it's your call for convenience xox

Nubbie:
See many do write up the ID in the description which makes the checkbox ticking a bit easy, still, maybe could re-write stuff to this PR as it was mostly intended to show what PR they apply and there is some unnecessary text in here xox

Gymnasiast:
Are you happy with the PR as it is now? I'm ok with merging this.

Nubbie:
@Gymnasiast If nothing else is needed and this PR is 'needed' - Yes 😄

Issue #723, opened 2016-06-07
STR_5873 - STR_5874


octaroot:
New strings from PR OpenRCT2/OpenRCT2#3818.

```
STR_5873    :Allow chain lifts on all track pieces
STR_5874    :{SMALLFONT}{BLACK}Allows any piece of track to be made into a chain lift
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] sv-SE
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] it-IT

PR #743, opened 2016-06-12
it-IT: Fixed translation errors, added strings

LucaRed:
Fixed a lot of translation errors. Added some missing strings


LucaRed:
Sorry, I fixed #723 but I forgot to mention it in the second commit

Issue #716, opened 2016-06-06
 STR_5871 - STR_5872


octaroot:
New strings from commit OpenRCT2/OpenRCT2@809e45949537ac533d11ed82fd186a01ade6b97d.

```
STR_5871    :Plants don't age.
STR_5872    :{SMALLFONT}{BLACK}Disable plant aging such that they don't wilt.
```

Updated in:
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] sv-SE
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] it-IT

Issue #730, opened 2016-06-08
STR_5875 - STR_5879


Nubbie:
New strings from PR OpenRCT2/OpenRCT2#3803
```
STR_5875    :Drawing Engine:
STR_5876    :{SMALLFONT}{BLACK}The engine to use for drawing the game.
STR_5877    :Software
STR_5878    :Software (hardware display)
STR_5879    :OpenGL
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] sv-SE
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] it-IT

PR #745, opened 2016-06-13
Update pt-BR.txt

renansimoes:
new strings and translation


LucaRed:
This fixes the issues #716, #723 and #730 for pt-BR

Wirlie:
@LucaRed "Fixes/Fix" is a correct term, but you would can use "Apply".
We usually use this term: (for example: Apply #XXX)
sin titulo

PS: This is a suggestion :)

Issue #806, opened 2016-07-16
STR_5897


octaroot:
New strings from PR OpenRCT2/OpenRCT2#3805.

```
STR_5897    :Window limit:
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW

PR #810, opened 2016-07-17
FR: Apply #806 & #816

rvgr:
Apply #806 & #816

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -4207,6 +4207,7 @@ STR_5894    :Entrez le taux de change
 STR_5895    :Sauvegarder la voie
 STR_5896    :La sauvegarde de la voie a échoué !
 STR_5897    :Nombre de fenêtres max. :
+STR_5898    :{BLACK}Impossible de charger la voie, le fichier est peut-être corrompu ou manquant!
 
 #############
 # Scenarios #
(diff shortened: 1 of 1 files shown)


Nubbie:
See you shorten it to make it fix the window before the box
If it could be re-phrased (to get rid of the dot), that would be appreciated

If not, Okey for me :+1:

rvgr:
Well it can hardly be rephrased, a more concise, literal translation of the string wouldn't mean anything in French. Anyway this abbreviation is very common and has already been used throughout the translation (including the original translation), so I don't think it will be a problem.

Nubbie:
Then It's a OK from me :+1: 
The biggest concern was the '. :' 
Was a bit 'messy'

rvgr:
Yes I see, nothing to worry about though, that's quite common!

Nubbie:
OK from me :+1:

Nubbie:
Have reviewed the new line you added, could you please insert a {NEWLINE} to break the line just like it does in the en-UK? 
Makes it easier to read in-game

Right now, it breaks at 'le fichier' which is one letter too much

image

rvgr:
I had to add two {NEWLINE} in order to wrap the string correctly. I preferred how it wrapped without it though.

Nubbie:
Presume that a {NEWLINE} isn't needed if you have to use two for it too wrap right, just ends up with too many lines :disappointed: 

You can revert it like it was, when reverted;
OK from me :+1:

Nubbie:
OK from me :+1:

PR #815, opened 2016-07-18
es-ES: Apply #806, #816

Wirlie:
#806 #816


Nubbie:
Looks good to me :+1:

PR #824, opened 2016-07-22
[example] Remove strings that match en-GB

marijnvdwerf:
An automated attempt for what's currently being discussed on gitter


runelaenen:
What about strings that are just the same (at the moment) for two languages. Changing a certain ingame string would become more difficult because you can't just ctrl+F the string itself anymore in the foreign (= not english) language.

Nubbie:
Removing the strings would leave a ton of gaps that would really take a longer time to fill in as it might be spread out everywhere, also, wouldn't this ruin 'a status bar' of untranslated strings?

Goddesen:
I agree with the comments of both @runelaenen and @Nubbie. Some words are the same in other languages, but they are not "duplicates" per say - they are indeed translated. Having them inherited from the English version is 1) confusing, requiring lookup in two files at once, and 2) not necessarily correct: Some strings, like "OpenGL" will always be the same, they can be inherited, but some words might just coincidentally be the same as the English one.

As @Nubbie mentions, there will be massive amounts of whitespace where things have not been translated yet - in the Norwegian translation, there are a lot of areas where only some strings have been translated as I would search through the document for repeats of a sentence or word I was translating (like "rollercoaster" -> "berg-og-dal-bane", one hell of a word to type repeatedly).

danidoedel:
I really don't like this as I agree with you guys... :/

IntelOrca:
langcheck can not tell whether a string is missing because it should be the same in both language or because it hasn't been translated yet. This is why I suggested only removing untranslated strings by commenting them out or using something like `{INHERIT}` to signify that the string is translated, but is the same as its sub language.

PR #826, opened 2016-07-22
Experiment #1 - Russian strings, added untranslated strings hached

Nubbie:
NOTE; Have added a new description in the bottom


Gymnasiast:
The problem is that the hashed strings will not reflect any updates made later. Referring to the en_GB translation is better. I'm therefore closing this PR.

PR #827, opened 2016-07-22
Experiment #2 - User friendliness with hashed strings

Nubbie:
A bit of a test for @Goddesen which translates the game from time to time, a check to see how practical it's for translators to remove the hashtags used for langcheck


Gymnasiast:
Closing this for the same reason as the other: hashed strings will go out of date. An untranslated up-to-date English string is still better than a translated string that was translated for an out-of-date hashed English string.

Issue #935, opened 2016-11-21
String 5879 (OpenGL labelled experimental)


Gymnasiast:
Changed string from PR OpenRCT2/OpenRCT2#4716.

```
STR_5879    :OpenGL (experimental)
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW

Nubbie:
[ sv-SE have applied ](https://github.com/OpenRCT2/Localisation/commit/cac4820f8f8e36220e62b696474c9bfbdc0928d6)

PR #944, opened 2016-11-24
Apply 935

Nubbie:
If you're applying strings from a issue, please fill it out below (if not, delete this text)

- Applying for issue: #935


Gymnasiast:
There are many things wrong with this:
- You put the string at the bottom instead of in order
- The string is defined twice, once at the correct place and once at the wrong place
- The PR violates the guidelines for the commit message and PR description that you yourself thought up(!)

PR #986, opened 2017-02-02
Experimental: ru-RU comment out strings & make translation-friendly

Nubbie:
This is a "test" to make the language checker more accurate **AND** help new translators

--------------
Experiment 1:
~~### What have changed;~~
~~* Remove all `<removed string> & comments inbetween strings ` **¤**~~
~~* Hash out all untranslated strings **¤¤**~~
~~* Re-making beginning & end comment ~~
~~* Refactor to make it more friendly for beginners **¤¤¤**~~
~~* Instructions~~

~~¤ _Only there for the original uk file, not needed for translators, distraction_~~
~~¤¤ _Manually done and "manually checked twice"_~~
~~¤¤¤ _The text was more made for programmers than translators, re-written for translating_~~

--------------
Experiment 2:
### 
* Instructions
* Re-making beginning & end comment
* Remove all `<removed string> & comments inbetween strings
* Remove all English strings & save chunks off translated text
* Place in a 'new function' where it tells what strings are missing 

Changes the size off the file, makes the statuses slightly more accurate, saves all translated strings

--------------

Should be accepted by @Gymnasiast @IntelOrca at least if not @marijnvdwerf have any thoughts

### DIFF
--- a/data/language/ru-RU.txt
+++ b/data/language/ru-RU.txt
@@ -1,134 +1,3469 @@
-# STR_XXXX part is read and XXXX becomes the string id number.
-# Everything after the colon and before the new line will be saved as the string.
-# Use # at the beginning of a line to leave a comment.
-STR_0000    :
-STR_0001    :{STRINGID} {COMMA16}
-STR_0002    :Спиральные горки
-STR_0003    :Стоячие горки
-STR_0004    :Подвесные раскачивающиеся горки
-STR_0005    :Обратные горки
-STR_0006    :Детские горки
-STR_0007    :Миниатюрная железная дорога
-STR_0008    :Монорельс
-STR_0009    :Мини стоячие горки
-STR_0010    :Заезд на лодках
-STR_0011    :Деревянная дикая мышь
-STR_0012    :Steeplechase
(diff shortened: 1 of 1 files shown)


Gymnasiast:
I have said it before: I don't want untranslated strings in language files, commented out or not. There is a high risk of translators translating obsolete strings or strings that have been changed in the meantime.

Nubbie:
Fair point
Should all commented out strings be removed instead and with a gap left in the string order to see where there's something missing or another proposal?
Like ignoring it all? xox

Gymnasiast:
I suppose you could leave just the identifiers (STR_0001) if that helps, but I'm not sure it's worth it.

Nubbie:
The only real thing I can come up with is clean up the language file / remove chunks where larger chunks off text is not translated :v

Gymnasiast:
I have updated the Russian file with the Russian GOG translation.

As for improving the process: I think moving to a specialised service or using an existing system (like gettext) would be better than attempting to fix it with comments and things like that. Thanks for your effort in any case. 👍

PR #1006, opened 2017-02-24
Readme: Add new topic, add new language ca-ES and move up nb-NO to maintained

Nubbie:
Just some slight maintenance as a new language were added

Note; "Checking & testing strings in-game" could be written differently and better


Nubbie:
**Optional text:**

---------

### Trying out strings in-game
You can test the translated strings in-game before publishing a pull request which is a way to check the string so it doesn't go out of boundaries/break at a wrong position if it contains a ``<newline>``.<br/>
One simple way to test the new language is to copy the whole document you've translated, navigate to OpenRCT's installation folder and then the ``\OpenRCT2\data\language`` folder and paste the copied text into the language you want to test.

PR #1038, opened 2017-04-03
Format & status fix (broken)

Nubbie:
The status bars are stuck at 0% - this PR doesn't fix that
but the bars are also in a cluster, the formating has been corrected and the invisible comment between the bars have been removed to restore their shape


Nubbie:
Ah, presume the error is what cause the bars to show 0% :u

Issue #1057, opened 2017-04-16
String 6039 - 6041


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#5368 and others.

```
STR_6039    :{SMALLFONT}{BLACK}Quick demolish ride
STR_6040    :Edit Scenario Options
STR_6041    :{BLACK}No mechanics are hired! 
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

telk5093:
Check for ko-KR plz.

Nubbie:
@rvgr  fr-FR is the only language left :)

JoelTroch:
@Gymnasiast @Nubbie The french translation is already up to date, you can check the box.

PR #1058, opened 2017-04-16
Update es-ES.txt String 6039 - 6041

siomy11:
If you are applying translations for an issue (or multiple), please list them below (if not, delete this text)

- Applying for issue: #1057


Wirlie:
hum... I do not agree with the translation that you have used for the string 6039 "{SMALLFONT}{BLACK}Quick demolish ride"

"Demoler atracción rápidamente" is the appropriate translation.

PR #1062, opened 2017-04-17
Update fr-FR.txt

siomy11:
If you are applying translations for an issue (or multiple), please list them below (if not, delete this text)

- Applying for issue: #1057

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -4347,6 +4347,9 @@ STR_6035    :Veuillez choisir votre installation de RCT1
 STR_6036    :{SMALLFONT}{BLACK}Vider
 STR_6037    :Veuillez choisir une installation valide de RCT1
 STR_6038    :{SMALLFONT}{BLACK}Si vous avez une installation de RCT1, définissez cette option pour pouvoir charger les scénarios, musiques...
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Tagging @JoelTroch .

@siomy11 :
Are you a native (or near-native) speaker?

If not, we have a translator for French who is.

JoelTroch:
The quality of this translation is good in overall but before it gets merged I would like to see the following changed:

"manège" is male, not female, so you would use "le" rather than "la".

"méchannique" is wrong, maybe you tried "mécanique" which means the domain but the person who repair mechanical stuff is "mécanicien" for a man or "mécanicienne" for a woman.

"à embauché" is almost correct, like this it means you "hired someone" (past action), it would be better to use "à embaucher" instead.

And final very minor remark, we frenchies have a weird convention of having a space before some type of punctuation, this applies to exclamation (!) and interrogation marks (?), so the final result would be "à embaucher !"

Otherwise, LGTM

siomy11:
@JoelTroch @Gymnasiast seems like I need more practice! Will fix it right away 👍

JoelTroch:
@Gymnasiast Now it's good for the merge ^^

Issue #1072, opened 2017-04-25
Modified strings 2686-2693, 3211 and new strings 6042-6054


Broxzier:
New strings from PR OpenRCT2/OpenRCT2#5282.

```
STR_6042    :Load height map
STR_6043    :Select height map
STR_6044    :Smooth height map
STR_6045    :Strength
STR_6046    :Normalise height map
STR_6047    :Smooth tiles
STR_6048    :Height map error
STR_6049    :Error reading PNG
STR_6050    :Error reading bitmap
STR_6051    :The width and height need to match
STR_6052    :The heightmap is too big, and will be cut off
STR_6053    :The heightmap cannot be normalised
STR_6054    :Only 24-bit bitmaps are supported
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

renansimoes:
These strings are already translated in pt-br. Please verify.

PR #1076, opened 2017-04-26
Update es-ES.txt

siomy11:

Applying for issue(s):
- #1072 

Might be wrong with the 'height' translation, since I feel like it could also be 'nivel' but 'altura' makes sense to me. 
@Wirlie Please fix if wrong! Thank you! :)


Wirlie:
Thank you for your collaboration, all translations are correct, but I have a suggestion with the translation of the string 6051, "El ancho y la altura deben coincidir" makes sense to me, what do you think of that? Anyway, your translation is correct :P

siomy11:
You're right ! 
La verdad no me acordaba como decir width en español Hahah 
Will change it soon, thanks!

Wirlie:
It's ok now :ok_hand: :3 

 PD: @siomy11 ¿Usas gitter? Podemos ponernos en contacto por ahí, por si hay alguna duda :p

siomy11:
@Wirlie No uso gitter, pero crearé una cuenta 👍 👍

Issue #1111, opened 2017-06-08
Strings 6101-6102


telk5093:
New strings from PR OpenRCT2/OpenRCT2#5196.

```

STR_6101    :Rides don't decrease in value over time
STR_6102    :{SMALLFONT}{BLACK}The value of a ride won't decrease over time, so guests will not suddenly think that a ride is too expensive.
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

Nubbie:
@JoelTroch FR have not applied

JoelTroch:
@Nubbie I'll try to find some free time to handle this. Thanks for the reminder ^^

Issue #1108, opened 2017-06-07
String 6064 to 6100


Gymnasiast:
New strings from PRs:
- OpenRCT2/OpenRCT2#5056
- OpenRCT2/OpenRCT2#5557

```
STR_6064    :Own all land
STR_6065    :Log user actions
STR_6066    :{SMALLFONT}{BLACK}Logs all user actions to files in your user directory.
STR_6067    :Server started.
STR_6068    :Server shutdown.
STR_6069    :{STRING} was kicked from the server by {STRING}.
STR_6070    :{STRING} was set to group '{STRING}' by {STRING}.
STR_6071    :{STRING} created new player group '{STRING}'.
STR_6072    :{STRING} deleted player group '{String}'.
STR_6073    :{STRING} edited permissions for player group '{String}'.
STR_6074    :{STRING} changed player group name from '{String}' to '{String}'.
STR_6075    :{STRING} changed the default player group to '{String}'.
STR_6076    :{STRING} used/toggled cheat '{STRING}'.
STR_6077    :Add Money
STR_6078    :{STRING} created ride '{STRING}'.
STR_6079    :{STRING} demolished ride '{STRING}'.
STR_6080    :{STRING} changed the appearance of ride '{STRING}'.
STR_6081    :{STRING} changed the status of ride '{STRING}' to closed.
STR_6082    :{STRING} changed the status of ride '{STRING}' to open.
STR_6083    :{STRING} changed the status of ride '{STRING}' to testing.
STR_6084    :{STRING} changed the vehicle settings of ride '{STRING}'.
STR_6085    :{STRING} changed the ride settings of ride '{STRING}'.
STR_6086    :{STRING} renamed the ride '{STRING}' to '{STRING}'.
STR_6087    :{STRING} changed the price of ride '{STRING}' to {STRING}
STR_6088    :{STRING} changed the secondary price of ride '{STRING}' to {STRING}
STR_6089    :{STRING} renamed the park from '{STRING}' to '{STRING}'.
STR_6090    :{STRING} opened the park.
STR_6091    :{STRING} closed the park.
STR_6092    :{STRING} changed the park entrance fee to {STRING}
STR_6093    :{STRING} placed new scenery.
STR_6094    :{STRING} removed scenery.
STR_6095    :{STRING} edited scenery.
STR_6096    :{STRING} set sign name to '{STRING}'.
STR_6097    :{STRING} placed a track of ride '{STRING}'.
[...]

Issue #1105, opened 2017-06-06
Strings 6059-6063


telk5093:
New strings from PR OpenRCT2/OpenRCT2#5133 and OpenRCT2/OpenRCT2##5307.

```
STR_6059    :{RIGHTGUILLEMET}
STR_6060    :Show guest purchases as animation
STR_6061    :{SMALLFONT}{BLACK}Show animated money effect{NEWLINE}when guests make purchases.
STR_6062    :{OUTLINE}{GREEN}+ {CURRENCY2DP}
STR_6063    :{OUTLINE}{RED}- {CURRENCY2DP}
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

Nubbie:
@Goddesen

Goddesen:
@Nubbie I don't think 52 days is too bad? 😄 
I was hoping I could get to this quickly after I was notified, but it seems that won't be the case just yet. But now I'm aware!

Nubbie:
@Goddesen You've had a longer streak ;)
Hopefully there will be a bot mentioning people, like it were said on Gitter :p

PR #1113, opened 2017-06-11
it-IT: apply #1105, #1108, #1111 and other enhancements

LucaRed:
Applying for issue(s):
- #1105 
- #1108 
- #1111


Gymnasiast:
I see the Italian translation is missing some strings in the region of STR_0512 up to and including STR_0602. See [here](https://github.com/OpenRCT2/OpenRCT2/blob/develop/data/language/en-GB.txt#L516-L606). Could you add the missing strings in another pull request?

PR #1128, opened 2017-06-23
All languages: Adding missing string 6034

Nubbie:
To correct the "99%" the language checker pushes out on some languages
PS. Wish I could squash them all, but my Github and localisation folder ain't getting along :(


LucaRed:
I've always wondered why it never reached 100%, nice catch! :thumbsup:

Gymnasiast:
Yeah, good catch, not sure how I missed it for Dutch.

Don't worry about squashing, I can do that with one press on a button.

Nubbie:
Wasn't too hard to find - The output on AppVayor say which string is missing 😄

PR #1159, opened 2017-07-23
ko-KR: Fix mistranslation and Apply #1162

telk5093:
Little question: Do you have any way to switch those two {STRING}s?

* Edit:
Add applying #1162 and just merge this plz (without clearing the problem above I wrote).


Gymnasiast:
@IntelOrca

PR #1166, opened 2017-07-24
arabic lang

Omaranwa:
hello ::
its arabic lang for the game i will update it as i can

### DIFF
--- a/data/language/ar-ُEG.txt
+++ b/data/language/ar-ُEG.txt
@@ -23,26 +23,26 @@ STR_0018    :Dinghy Slide
 STR_0019    :Mine Train Coaster
 STR_0020    :Chairlift
 STR_0021    :Corkscrew Roller Coaster
-STR_0022    :Maze
+STR_0022    :متاهة
 STR_0023    :Spiral Slide
 STR_0024    :Go Karts
 STR_0025    :Log Flume
 STR_0026    :River Rapids
 STR_0027    :Dodgems
-STR_0028    :Pirate Ship
+STR_0028    :سفينة قراصنة
 STR_0029    :Swinging Inverter Ship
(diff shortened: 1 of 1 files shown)


Gymnasiast:
ar-AR is not a valid code. It means 'Arabic from Argentina'.

I can look up the correct code for you. In which country do you live?

Nubbie:
As I only see one string translated, maybe you could make some more commits and changes before it being merged :) 
(Just so it gets updated and not left at <5%)

Omaranwa:
hello there i am from egypit. also as i sad  this is for testing only i will update it soon

Gymnasiast:
Good!

Could you rename the file to `ar-EG.txt`?

Omaranwa:
ok i will do

data/language/ar-ُEG.txt
```diff
@@ -3731,1103 +3731,4 @@ STR_5389    :{SMALLFONT}{BLACK}Skip to the selected command in the title sequenc
...
-STR_NAME    :Condor Ride
-STR_DESC    :Riding in special harnesses below the track, riders experience the feeling of flight as they swoop through the air in Condor-shaped trains
-STR_CPTY    :4 passengers per car
-
-#WW
-[CONGAEEL]
-STR_NAME    :Conger Eel Coaster
-STR_DESC    :A compact steel-tracked roller coaster where the Eel-shaped train travels through corkscrews and loops
-STR_CPTY    :4 passengers per car
+STR_5393    :{SMALLFONT}{BLACK}Remove the selected save from the
```
Gymnasiast: You cut off a large part of the file here. I think that was by accident.

Omaranwa: sorry for that

Issue #1168, opened 2017-07-25
Track designer strings


Gymnasiast:
Changed strings from PR OpenRCT2/OpenRCT2#5991.

```
STR_0888    :Quit Track Designer
STR_3207    :Track Designer
STR_3335    :Track Designer - Select Ride Types & Vehicles
STR_3344    :Track Designer
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

JoelTroch:
@Gymnasiast Already done for fr-FR

Nubbie:
[ zhTW have applied ](https://github.com/OpenRCT2/Localisation/commit/f702d42719b119e40bbaba4979a79ce3fbf97869#diff-836f39ea934cec0e858c8d9ddf701d75R893)

Gymnasiast:
@J0anJosep

Gymnasiast:
(Not actually applied in Catalan yet.)

PR #1169, opened 2017-07-25
Sv-SE: Apply #1168

Nubbie:

Applying for issue(s):
 #1168

### DIFF
--- a/data/language/sv-SE.txt
+++ b/data/language/sv-SE.txt
@@ -890,7 +890,7 @@ STR_0884    :Ladda Landskap
 STR_0885    :Spara Landskap
 STR_0886    :Avsluta Spel
 STR_0887    :Avsluta Scenarioskaparen
-STR_0888    :Avsluta Berg- och Dalbanebyggaren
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Why do you keep filling in the form incorrectly? You created most of it!

Also, the commit message is wrong again: it should be `sv-SE: Apply #1168`, not `Apply 1168`.

I really don't understand you: you came up with so many rules and templates, yet you frequently violate them.

Nubbie:
I don't see how I fill it out incorrectly, is it because I don't delete the first part? :s 
The only rule & template I came up with, regarding your comment, is the one to connect and easily mark status on PR, just for the purpose to link the PR and Issue, see picture 
kpnjfgeawpkajl

If the hashtag marking is on the commit, the commit and issue gets linked (which makes this marking bellow), harder too see but links the commit instead (commit and issue)
ljiokgeasioj

The only thing/rule I can come up with making is the templates for readme, pr and issue to help and guide new people and help maintainers to see which has complied the issue/who maintainers are

PR #1172, opened 2017-07-26
Add some basic Esperanto (eo) translations

flipflop97:


Nubbie:
Hey and welcome :)
You should read the ["Creating a new language"](https://github.com/OpenRCT2/Localisation#creating-a-new-language) chapter in the read me
You use the wrong ISO-codes for the filename 
As most parts are untranslated, I suggest you make more edits and translate more parts before being added to the game :)

flipflop97:
Yes I read that. The problem with Esperanto is that it is not connected to a country. The official ISO-code therefore has no country suffix, leaving it `eo`.

I'll submit it again later when I have translated more.

Gymnasiast:
I'm willing to accept a translation for a constructed language like Esperanto, but only if you maintain it well. For natural languages I'm more willing to accept to risk of it becoming outdated, but this won't really make the game more accessible for more people. I therefore set the bar a bit higher.

So, simply put, do you think you can update the translation regularly? Once a week will suffice.

The ISO code `eo` is OK, as Esperanto is not tied to any country, as you already pointed out.

flipflop97:
Hmm, I think i'll pass for now then. I'm not an Esperanto expert, so I don't know if I'm capable of fully and regularly maintaining it.

Gymnasiast:
Ok, feel free to come back if things change. :-)

PR #1202, opened 2017-08-20
en_US: Use "cattails" instead of "bulrushes"

Keatzee:
Pretty trivial, I know... but "cattail" is far more common in American English. In fact, I had never
even heard of "bulrushes" before noticing in-game.

Plus the USDA uses "cattail" as well, so I guess it's official
https://plants.usda.gov/core/profile?symbol=TYPHA


Gymnasiast:
Oh, don't worry about triviality. We have separate American and British English translations for a good reason. I have replaced some Britishisms in the American translation as well.

Keatzee:
Good to know! I'll keep an eye out for any more I can find

Issue #1217, opened 2017-09-13
Strings 6145 - 6147


Gymnasiast:
New strings from PRs OpenRCT2/OpenRCT2#6272 and OpenRCT2/OpenRCT2#6273.

```
STR_6145    :{SMALLFONT}{BLACK}Set speed limit for boosters
STR_6146    :Enable all drawable track pieces
STR_6147    :{SMALLFONT}{BLACK}Enables all track pieces the ride type is capable of in the construction window, regardless of whether the vehicle supports them.

```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES

Nubbie:
Cant find the changes  OpenRCT2/OpenRCT2#6273 did, context?

Omaranwa:
add it to arabic as well>

Nubbie:
[zh-CN have applied](https://github.com/OpenRCT2/Localisation/blob/master/data/language/zh-CN.txt#L4463-L4465)

PR #1223, opened 2017-09-19
Update ar-EG.txt

Omaranwa:

Applying for issue(s):
- #1217


Gymnasiast:
Issues like #1217 are there to notify you there are new strings you should translate (in your case, into Arabic). You do not need to add English strings to your translation. If you don't specify a certain string, it will fall back to English automatically.

Omaranwa:
i am gona translate soon you have to wait

PR #1240, opened 2017-10-05
ja-JP: add translations for more strings; unify more ride strings.

AaronVanGeffen:
With the new font support on its way, I've picked up where we left off. This adds a bunch of translations and unifies some existing strings a little.

For ride names, I've used the RCT3 fan-translation for reference (e.g. [this page](https://www39.atwiki.jp/rct3jpinfo/pages/81.html).

### DIFF
--- a/data/language/ja-JP.txt
+++ b/data/language/ja-JP.txt
@@ -5,22 +5,22 @@ STR_0000    :
 STR_0001    :{STRINGID} {COMMA16}
 STR_0002    :スパイラルコースター
 STR_0003    :スタンディングコースター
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
2e9983c810163f7c831c3a47354fc52621d723b0 updates the untranslated strings with those from the British English file. This may seem a bit tedious, but my side-by-side diff tool was going haywire over the latter half of the file, so this makes my life a little easier. Moreover, obsolete strings are now marked as such.

Gymnasiast:
I'm a bit apprehensive about adding English strings. But if it helps you do a translation quicker, I'm not going to object.

Issue #1259, opened 2017-10-15
String 6154+6155


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#6438.

```
STR_6154    :It is not recommended to run OpenRCT2 with elevated permissions.
STR_6155    :Neither KDialog nor Zenity are installed. Please install one, or configure from the command line.
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP

Notification when new strings arrive:
@Nubbie @Goddesen

telk5093:
ko-KR is applied

Issue #1257, opened 2017-10-14
STR_6156


trigger-death:
New strings from PR OpenRCT2/OpenRCT2#6471.

```
STR_6156    :Name is reserved
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP

Notification when new strings arrive:
@Nubbie @Goddesen

telk5093:
``STR_6156    :Name is reserved``

Nubbie:
It hasn't been merged yet
And the string is incorrect, see telk5093's comment

Gymnasiast:
Closing this until the relevant PR is merged.

PR #1264, opened 2017-10-16
cs-CZ: Apply #1257 #1259

octaroot:
cs-CZ: Apply #1257 #1259


Gymnasiast:
One of the first, as usual ^.^

PR #1266, opened 2017-10-16
ja-JP: advertisement campaigns, view options, misc.

AaronVanGeffen:
Translations for advertisement campaigns, view options, misc.

I've also looked into transcribing the scenario names in katakana. This is a work in progress, and as such not committed yet. It would be useful to have a file with the original English descriptions. These exist for the scenarios from RCT1 in en-GB.txt. Is a similar listing available somewhere for RCT2?

As SV6 limits park names to 32 bytes, some of these transcriptions will have their name chopped off by a few characters. Their names in the scenario selection screen appear just fine. Would this be acceptable, or would you recommend I just transcribe STR_SCNR and leave STR_PARK to the English name in such cases?


Gymnasiast:
Perhaps better to wait with the park names until the SV6 replacement comes along? You _can_ do the ones from the scenario screen if you feel katakana would fit better with Japanese expectations.

PR #1280, opened 2017-10-22
ja-JP: scenario objectives; some multiplayer terminology; misc

AaronVanGeffen:
This PR finally tackles scenario objectives, as well as some multiplayer terminology and other miscellaneous strings.

Also: スタフ→スタッフ. Apparently, the latter is a more common transcription for 'staff'.


data/language/ja-JP.txt
```diff
@@ -4089,10 +4089,10 @@ STR_5778    :建設: {COMMA16}年前
 STR_5779    :収入: {CURRENCY2DP} 時あり
 STR_5780    :維持費: {CURRENCY2DP} 時あり
 STR_5781    :維持費: 未知
-STR_5782    :You are now connected. Press '{STRING}' to chat.
+STR_5782    :今から接続中です。「{STRING」でチャットできます。
```
Gymnasiast: You accidentally removed the right brace.

AaronVanGeffen: Oof, nice catch!

PR #1281, opened 2017-10-23
ko-KR: Change date format order

telk5093:
Change date format order from {date}/{month} to {month}/{date} in that way used in korea.


Gymnasiast:
We have a setting for that. Does it not work?

AaronVanGeffen:
The setting just changes the date as displayed in the bottom-right corner, as far as I know. There are different strings for date formatting, in this case used to display the research completion date. Another is used to display scenario completion dates, etcetera.

Ideally, those would be covered by the setting, but they are not. In the mean time, I believe this is an appropriate solution.

Gymnasiast:
Ok, I will merge it then.

It's probably better to remove that setting and just put everything in the translation file. There's little point in it anyway - no sane Dutch player would want their dates displayed as 'maart 2', for example. I'll get the gun.

Gymnasiast:
@telk5093 What is the Korean format if you include the year? (Like 1st March, Year 1)

telk5093:
@Gymnasiast It is {year}/{month}/{date}, if you meant the order (ex. 2017년 10월 24일)
(년=year, 월=month, 일=day)

Gymnasiast:
@telk5093 Yup, that's what I meant.

telk5093:
So I issued OpenRCT2/OpenRCT2#2313, but there is no way to change the order of translation strings easily, except for {POP16} and {PUSH16} as you know.
My thought is the same with @Gymnasiast in that no one wants to see wrong ordered date format in their language

AaronVanGeffen:
Yes, proper format strings would be an ideal solution over popping and pushing on the stack… As far as I know, that hasn't been in the works yet, though.

Issue #1285, opened 2017-11-02
STR_6158


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#6584.

```
STR_6158    :Failed to load this file...{NEWLINE}Incompatible RCTC version: {COMMA16}
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP
- [x] ar-EG

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot

Gymnasiast:
Quick as always ;-)

Gymnasiast:
@danidoedel @Goddesen

PR #1286, opened 2017-11-02
cs-CZ: Apply #1285

octaroot:
cs-CZ: Apply #1285


Gymnasiast:
Oh, you're even quicker than usual! :o

octaroot:
What can I say, I like the game too much :) also - it's because of the GitHub notifications

PR #1306, opened 2017-11-22
ko-KR: Support OpenRCT2/OpenRCT2#6710

telk5093:
Support OpenRCT2/OpenRCT2#6710, representing right ride vehicle names.


Gymnasiast:
This should not be needed. It should fall back on STR_NAME. But I just checked and saw that it doesn't. I'll fix this.

telk5093:
@Gymnasiast Okay, thank you

Gymnasiast:
The issue is now solved in `develop`.

PR #1408, opened 2018-01-22
cs-CZ: slight improvements

Aroidzap:
Minor changes to some words.

### DIFF
--- a/data/language/cs-CZ.txt
+++ b/data/language/cs-CZ.txt
@@ -17,8 +17,8 @@ STR_0012    :Dostihy
 STR_0013    :Autodráha
 STR_0014    :Volný pád
 STR_0015    :Bobová dráha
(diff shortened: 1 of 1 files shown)


Gymnasiast:
@octaroot 

Discussion can take place in Czech.

octaroot:
Hi @Aroidzap, thanks for your new contribution ;-)

I reviewed your changes and they are all fine by me.

The only thing I'm not 100% sure about is reverting back to the untranslated term _looping_ (in favor of _lopink_). The dictionary I used listed both, so I opted for the more "Czech" translation. However since both are valid, feel free to leave it in - if you think it's more fitting.

Aroidzap:
@octaroot I think that these taken over words are used more widely in untranslated forms.

octaroot:
OK, ready to merge @Gymnasiast

Issue #1411, opened 2018-01-29
Translations for vehicle objects


Gymnasiast:
**If any of these entries are already present, update them. Do not remove fields if they are already in your translation, only change or add what's below here.**

```
#####################
# Rides/attractions #
#####################

#RCT2
[BMSD]
STR_NAME    :Twister Trains
STR_DESC    :A spacious train with shoulder restraints

[BMSU]
STR_NAME    :Stand-up Twister Trains
STR_DESC    :A train with shoulder restraints, in which the riders stand up

[BMFL]
STR_NAME    :Floorless Twister Trains
STR_DESC    :A spacious train with shoulder restraints and no floor, making for a more exciting ride

[WMMINE]
STR_NAME    :Mine Cars
STR_DESC    :Cars shaped like an old mine cart

[WMOUSE]
STR_NAME    :Mouse Cars
STR_DESC    :Indivual cars shaped like a mouse

[STEEP1]
STR_NAME    :Horses
STR_DESC    :Single cars shaped like a horse

[STEEP2]
STR_NAME    :Motorbikes
STR_DESC    :Single cars shaped like a motorbike

[SBOX]
STR_NAME    :Soap boxes
STR_DESC    :Single cars shaped like a soap box

[BOB1]
STR_NAME    :Bobsleigh Trains
STR_DESC    :A train consisting of 2-seater cars where the riders are behind each other

[INTBOB]
STR_NAME    :6-seater Bobsleighs
STR_DESC    :Bobsleighs with three seating rows, with room for two people on each

[LIFT1]
STR_NAME    :Lift Cabin
STR_DESC    :Steel lift cabin

[ARRT1]
STR_NAME    :Corkscrew Roller Coaster Trains
STR_DESC    :Roller coaster train with shoulder restraints

[ARRT2]
STR_NAME    :Hypercoaster Trains
STR_DESC    :Comfortable trains with only lap bar restraints

[MONBK]
STR_NAME    :Bicycles

[OBS1]
STR_NAME    :Single-deck Cabin

[OBS2]
STR_NAME    :Double-deck Cabin

[GTC]
STR_NAME    :Ghost Train Cars

[HMCAR]
STR_NAME    :Haunted Mansion Cars

[VREEL]
STR_NAME    :Virginia Reel tubs

[WMSPIN]
STR_NAME    :Spinning Mouse Cars

[AMT1]
STR_NAME    :Mine Trains

[SLCT]
[...]

PR #1412, opened 2018-01-30
ko-KR: Apply #1411 and the Panda world

telk5093:
Applying for issue(s):
- #1411

### DIFF
--- a/data/language/ko-KR.txt
+++ b/data/language/ko-KR.txt
@@ -4524,6 +4524,7 @@ STR_6214    :마천루 B
 STR_6215    :건설
 STR_6216    :운영 및 운행
 STR_6217    :놀이기구 / 트랙 관련 기능
+STR_6218    :OpenRCT2 공식
 
 #############
 # Scenarios #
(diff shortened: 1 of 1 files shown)


data/language/ko-KR.txt
```diff
@@ -5066,9 +5065,7 @@ STR_DESC    :모형 골프 게임
 STR_CPTY    :
 
 [GTC]
-STR_NAME    :고스트 트레인
-STR_DESC    :으스스한 풍경과 특별한 효과 주변을 지나며 여러 높낮이를 오르내리는 괴물 모양의 차량
-STR_CPTY    :차량당 2명의 승객
+STR_NAME    :고스트 트레인 차량
```
Gymnasiast: Do not remove STR_DESC and STR_CPTY. These should stay. They're not listed because they're unchanged.

AaronVanGeffen: Just to clarify: they aren't listed in the English language file, as the strings from the original DAT files do not need to be adjusted.

telk5093: It might be better to notice me not to remove, anyway done. Please check it.

PR #1430, opened 2018-02-12
ko-KR: Change a translation

telk5093:
* Change a scenario name since it was a mistranslation
 * Change a scenario description since it was too long

And this scenario(South America - Lost City Inca) has no height restriction but scenario description says it has.
I've checked vanilla, there is also no height restriction.
Maybe it was a bug when this scenario was built.

For now, I deleted the phase "the height restriction is exists" to prevent confusion, but how do you think?


Gymnasiast:
Basically, there are two ways to solve this:
- Change the description
- Fix the scenario to include a height restriction.

I'll have to discuss this first.

PR #1436, opened 2018-02-18
Update ru-RU.txt Spell check

alex-0xFF:
Just a quick spell check for most obvious mistakes


Gymnasiast:
Thanks! Many of these errors were caused by the limited room that GOG had when translating RCT2 to Russian - the font they created therefore lacked the hard sign and e-with-diaeresis, which seems to be 90% of the problems that you corrected.

alex-0xFF:
Not at all. I'm very excited about this project, I will try to help as much as I can.

PR #1438, opened 2018-02-21
Update ru-RU.txt:  info: 3401 / 4518 (75%) strings present

alex-0xFF:


data/language/ru-RU.txt
```diff
@@ -3508,9 +3652,8 @@ STR_DESC    :Широкие вагончики падают по вертика
 STR_CPTY    :6 пассажиров
 
 [BOB1]
-STR_NAME    :Бобслей
-STR_DESC    : Люди едут в бобах по извивающемуся треку.
-STR_CPTY    :2 пассажира
+STR_NAME    :Бобслей-поезд
+STR_DESC    :Поезд состоящий из двухместных вагонов, где пассажиры сидят друг за другом
```
Gymnasiast: For all these entries: if Russian has STR_DESC or STR_CPTY but English has not, keep the Russian STR_DESC/STR_CPTY nevertheless. English does not need all of them, but Russian does.

alex-0xFF: How does it work internally? Just curious,

Gymnasiast:
> How does it work internally? Just curious,

Taking the following block as an example:
```
[BOB1]
STR_NAME    :Бобслей-поезд
STR_DESC    :Поезд состоящий из двухместных вагонов, где пассажиры сидят друг за другом
STR_CPTY    :2 пассажира
```

This is a block that overrides a description from a .DAT file. In RCT2, all vehicles, scenery, paths, and many more things are saved in .DAT files. These .DAT files also contains translations for a few languages. These include languages like UK English, Dutch and Korean, but not Russian.

OpenRCT2 tries to fetch translations from this .DAT file. If the .DAT file does not contain translations for a particular language (like Russian), it will fall back to UK English. Also, even if the .DAT file contains translation, they are not always correct. So in order to fully translate languages or to fix existing ones, we use a block like this.

In the case of `BOB1`, English only overrides STR_NAME and STR_DESC, because they are not correct, but it leaves STR_CPTY, which then falls back to the description in the .DAT file. Russian, however, has no descriptions in the .DAT file, so that will need STR_CPTY as well. Otherwise, the interface won't show "2 пассажира" but "2 passengers" in Russian.

Issue #1501, opened 2018-05-30
Strings 6202, 6256-6259


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#7577.

```diff
-STR_6202    :Enable virtual floor
+STR_6202    :Virtual floor style:
+STR_6256    :Rendering effects
+STR_6257    :Glassy (translucent)
+STR_6258    :Clear (transparent)
+STR_6259    :Disabled
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP
- [x] ar-EG

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel

PR #1502, opened 2018-05-30
ko-KR: Apply #1501

telk5093:
I thought that there is no proper simple translation for 'Glassy' or 'Clean' in korean, so I removed parentheses.

Applying for issue(s):
- #1501


Gymnasiast:
> I thought that there is no proper simple translation for 'Glassy' or 'Clean' in korean, so I removed parentheses.

That's ok, it's no problem if it doesn't exactly match the English strings. The most important thing is that it's clear to the user. The meaning is more important than the exact words that you use.

PR #1508, opened 2018-06-11
Use dingbats for +/- buttons in languages using TrueType fonts

AaronVanGeffen:
To be merged when https://github.com/OpenRCT2/OpenRCT2/pull/7664 has been merged.


Gymnasiast:
I'm still missing cs-CZ.

AaronVanGeffen:
Oh, I didn't realise it cs-CZ is also rendered with TrueType fonts…

PR #1524, opened 2018-07-23
de-DE: Use German quotation style

Gymnasiast:

### DIFF
--- a/data/language/de-DE.txt
+++ b/data/language/de-DE.txt
@@ -2751,7 +2751,7 @@ STR_2743    :Kopieren Sie „data/css17.dat“ aus Ihrem RCT1-Verzeichnis nach �
 STR_2744    :[
 STR_2745    :\
 STR_2746    :]
-STR_2747    :“
+STR_2747    :"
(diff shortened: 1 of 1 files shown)


danidoedel:
I wouldn't mix double quotes and single quotes. Just use double quotes for everything ;)
Also, you missed some strings like https://github.com/OpenRCT2/Localisation/pull/1524/files#diff-3841c99f5b7b3b48a6e056eacbe1fd97R2692

Gymnasiast:
Isn't the idea in German that you use double quotes for actual quotations (and thoughts) and single quotes for other things?

danidoedel:
No not really. I have never really used single quotes for anything, but you would use them in another quotation which is already using double quotes.

ccoors:
Why not „“ on STR_1080, STR_1081, STR_1085, STR_1086, STR_1087? I think this is wrong.

danidoedel:
Because it‘s easy to overlook them if 4 different styles were used. Of course they should be changed as well.

PR #1536, opened 2018-08-12
Fill-up blanks in French translation

Berbe:


data/language/fr-FR.txt
```diff
@@ -780,49 +780,49 @@ STR_0775    :Parc sans nom{POP16}{POP16}
...
-STR_0803    :25
-STR_0804    :26
-STR_0805    :27
-STR_0806    :28
-STR_0807    :29
-STR_0808    :30
-STR_0809    :31
-STR_0810    :janv
-STR_0811    :fév
+STR_0779    :1 ͤ ͬ
```
Gymnasiast: This is not going to work with OpenRCT2's sprite font.
Also, these are used for dates, and French only uses an ordinal number for the first of a month, and cardinal numbers for the rest.

PR #1556, opened 2018-09-11
Added new strings

DJHasis:
I used meld,compared to en_US and added the missing strings.


AaronVanGeffen:
It's not necessary to add strings that are identical to en-GB.

Gymnasiast:
**Never** compare with en_US. The source language is en_GB.

DJHasis:
Okay, I didn't know that. I'll compare to en_GB. But should I remove the deprecated strings also?

Gymnasiast:
Yup.

Gymnasiast:
Since there has been no reply and this PR only adds English strings, I'll close it.

Issue #1557, opened 2018-09-13
String 6266


AaronVanGeffen:
New string from PR OpenRCT2/OpenRCT2#7971.

```
STR_6266    :Open custom content folder
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP
- [x] ar-EG
- [x] da-DK
- [x] tr-TR

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel @LPSGizmo @BobbyS38

frenchiveruti:
I see, done.
Thanks


________________________________
De: Matte A <notifications@github.com>
Enviado: martes, 18 de septiembre de 2018 18:24
Para: OpenRCT2/Localisation
Cc: frenchiveruti; Mention
Asunto: Re: [OpenRCT2/Localisation] String 6266 (#1557)


@frenchiveruti<https://github.com/frenchiveruti> You edit this issue template to add yourself to the list by writing "@frenchiveruti<https://github.com/frenchiveruti>" at the end of the text
https://github.com/OpenRCT2/Localisation/blob/master/.github/ISSUE_TEMPLATE.md

Nubbie:
sv-SE have applied

PR #1558, opened 2018-09-14
es-ES: Added String 6266

frenchiveruti:
Added String.

Applying for issue(s):
- #1557


frenchiveruti:
Will "Abrir la carpeta de mods" suffice?
A translation would be "open mods folder".

Gymnasiast:
They're not mods.
(I didn't see you already changed it before I hit merge. Could you change it back to what it was initially and do another PR?)

PR #1565, opened 2018-09-15
Revert "es-ES: Added String 6266"

frenchiveruti:
Reverts OpenRCT2/Localisation#1558

### DIFF
--- a/data/language/es-ES.txt
+++ b/data/language/es-ES.txt
@@ -3733,7 +3733,7 @@ STR_6262    :Volúmen Maestro
 STR_6263    :{SMALLFONT}{BLACK}Activar o desactivar todos los sonidos
 STR_6264    :Usar siempre explorador de archivos del sistema
 STR_6265    :{SMALLFONT}{BLACK}Al activarse, el explorador de archivos del SO será usado en lugar del de OpenRCT2.
-
+STR_6266    :Carpeta de contenido personalizado
 
 ##############
 # Escenarios #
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
Any particular reason for this?

frenchiveruti:
> 
> 
> Any particular reason for this? Why not just change the string in question?

Yes, requested by @Gymnasiast on #1558

Wirlie:
"Ver carpeta cont. personalizado" would be an acceptable translation, but I do not have RCT2 on my computer at this moment, so I cannot test if the translation fits the dropdown.

frenchiveruti:
> 
> 
> "Ver carpeta cont. personalizado" may be an acceptable translation, but I do not have RCT2 on my computer at this moment, so I cannot test if the translation fits the dropdown.

From chat at gitter:
"What the button does is open the folder where all the content of the user resides. On Windows, this folder is C:\Users\Username\Documents\OpenRCT2. The user can then put their scenarios, saves, custom objects and other stuff in that folder. It's important that that meaning gets across - it's not important that it's a literal translation.

The character limit is hard to say, as it depends on the window width.

Perhaps we should go with the Spanish translation for "open content folder"."

Wirlie:
Oh i see, I really did not know the functionality of the dropdown button, so I've made a suggestion based on your first commit (https://github.com/OpenRCT2/Localisation/pull/1558/commits/57fc426762c61bd3d3de8a6de3a595330c7145dc), thank you for the clarification.

frenchiveruti:
Actually with a very ugly resolution ratio, the translation fits fine inside the box.
Game at 1176x611

Issue #1451, opened 2018-03-20
Strings 6226–6228


AaronVanGeffen:
New strings from PR OpenRCT2/OpenRCT2#7187.

```
STR_6226    :Enable early scenario completion
STR_6227    :{SMALLFONT}{BLACK}Triggers scenario completion when all scenario goals are met before the target date.
STR_6228    :Scenario Options
```

Please review your translation for the following changed string as well:
```diff
-STR_6170    :Tweaks
+STR_6170    :Interface Tweaks
```

Updated in:
- [x] en-US
- [x] cs-CZ
- [x] nl-NL
- [x] fr-FR
- [x] de-DE
- [x] pt-BR
- [x] es-ES
- [x] it-IT
- [x] sv-SE
- [x] nb-NO
- [x] ko-KR
- [x] zh-CN
- [x] zh-TW
- [x] ca-ES
- [x] ja-JP
- [x] ar-EG

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel

PR #1580, opened 2018-10-03
Modified STR_6170

pppy3des:
Applying for issue(s):
- #1451 

modified STR_6170


Gymnasiast:
Thanks for this update.

Since Japanese misses a lot of translations, perhaps you could take a look at adding them? It looks as if you're from Japan, and we don't have many native speakers, so we could really use your help.

PR #1582, opened 2018-10-10
Grammatical Change To US English

pizza2004:
In America I've never heard the phrase Is Good Value and rather always Is a Good Value.  This is related to OpenRCT2/OpenRCT2#8074 and OpenRCT2/OpenRCT2#8076.

### DIFF
--- a/data/language/en-US.txt
+++ b/data/language/en-US.txt
@@ -165,7 +165,7 @@ STR_1486    :{SMALLFONT}“I haven't finished my {STRINGID} yet”
 STR_1487    :{SMALLFONT}“Just looking at {STRINGID} makes me feel sick”
 STR_1488    :{SMALLFONT}“I'm not paying that much to go on {STRINGID}”
 STR_1489    :{SMALLFONT}“I want to go home”
-STR_1490    :{SMALLFONT}“{STRINGID} is really good value”
+STR_1490    :{SMALLFONT}“{STRINGID} is a really good value”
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Are you sure ”... is a really good value” is common in American English? It sounds odd to me. I wonder if this is how an American would say this.

pizza2004:
Well, I don't think we can ever really be /completely/ certain with something like language, but in addition to my personal experience I submit the following two things:
https://idioms.thefreedictionary.com/a+good+value
https://www.ldoceonline.com/dictionary/good-value-for-money

Notice the first /does/ use a in the idiom, meaning that's part of the phrase /somewhere/ in the world, and also notice that in the second link (I believe the same one you provided in the other issue) it /specifically/ states that it is British English that uses it that way.  These two things plus my own life experiences are enough for me to think that it's not proper US English.

pizza2004:
Actually enough, it's funny because from what I see online the phrase is "It's a good value for the money." in America, but it's "It's good value for money." in the UK.

Gymnasiast:
I don't doubt that "is really good value" is uncommon in American. What I'm talking about is its replacement - is "is a really good value" a common way of expressing it, or is there another phrase?

pizza2004:
Oh, then yes, definitely.  Born and raised in Utah all 25 years of my life I'll attest to that being what I actually hear people say.

tombomp:
I know this is about changing it in US English but just to confirm in UK English you would never put the a in and "a good value" sounds very strange to me.

PR #1614, opened 2018-12-11
Add missing languages to issue template

Umdlye:


AaronVanGeffen:
This might be a good opportunity to reorder the list alphabetically as well.

Umdlye:
Good one. Welcome back by the way!

PR #1668, opened 2019-03-10
Use trash in en-US.txt

confol1:
Changed litter to trash and changed litter bins to trash cans in en-US.txt.


Gymnasiast:
Am I right in assuming that "trash" is something that most Americans would use instead of "litter"?

confol1:
Yeah, everywhere that I've been to in America, people say trash and signs say trash or trash can.

Gymnasiast:
Ah good, that's what I thought (hence the remark on the previous PR).

Issue #1675, opened 2019-03-31
String 6304


telk5093:
New strings from PR OpenRCT2/OpenRCT2#8926.

```
STR_6304    :Open scenery picker
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel @LPSGizmo @BobbyS38 @frenchiveruti @anon569 @JoelTroch

J0anJosep:
ca-ES already done.

galfisk:
Applied to sv-SE

Issue #1670, opened 2019-03-25
Removed string 3194


Gymnasiast:
Removed string from PR OpenRCT2/OpenRCT2#8889.

```diff
-STR_3194    :Scenario Description
```

Updated in:
- [x] ar-EG
- [x] ca-ES
- [x] cs-CZ
- [x] da-DK
- [x] de-DE
- [x] en-US
- [x] es-ES
- [x] fr-FR
- [x] hu-HU
- [x] it-IT
- [x] ja-JP
- [x] ko-KR
- [x] nb-NO
- [x] nl-NL
- [x] pt-BR
- [x] sv-SE
- [x] tr-TR
- [x] zh-CN
- [x] zh-TW

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel @LPSGizmo @BobbyS38 @frenchiveruti @anon569 @JoelTroch

JoelTroch:
@Gymnasiast Please tick "fr-FR" as merged PR #1673 took care of this issue.

PR #1677, opened 2019-03-31
cs-cz apply #1675 #1670

octaroot:


Gymnasiast:
Thanks!

By the way, I hope you like the look of Czech with the RCT sprite font. I think it's a lot easier and nicer to read.

Issue #1691, opened 2019-05-01
Strings 1171, 1172, 1730, 2981, 6308


AaronVanGeffen:
Changed strings from PR OpenRCT2/OpenRCT2#8659:

```diff
-STR_1171    :{RED}Closed - - 
-STR_1172    :{YELLOW}{STRINGID} - - 
+STR_1171    :{RED}Closed
+STR_1172    :{YELLOW}{STRINGID}
```
```diff
-STR_1730    :{RED}Closed - - 
+STR_1730    :{RED}Closed
```
```diff
-STR_2981    :{RED}No entry - - 
+STR_2981    :{RED}No entry
```

New strings from PR OpenRCT2/OpenRCT2#8659:

```
STR_6308    :“{STRINGID}{OUTLINE}{TOPAZ}”{NEWLINE}{STRINGID}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel @LPSGizmo @BobbyS38 @frenchiveruti @anon569 @JoelTroch

telk5093:
New string
```
STR_6307    :Colour scheme: {BLACK}{STRINGID}
```
is missing (from OpenRCT2/OpenRCT2/pull/9128)

Gymnasiast:
@telk5093 I have opened a separate issue for that: #1697.

renansimoes:
This issue is already fixed in pt-br file, I don't know why isn't checked here.

J0anJosep:
ca-ES already done.

PR #1694, opened 2019-05-02
hu-HU: Update + Apply #1691

anon569:
Applying for issue(s):
- #1691


data/language/hu-HU.txt
```diff
@@ -1,4 +1,4 @@
-# STR_XXXX part is read and XXXX becomes the string id number.
+https://github.com/OpenRCT2/Localisation/issues/1691# STR_XXXX part is read and XXXX becomes the string id number.
```
AaronVanGeffen:
```suggestion
# STR_XXXX part is read and XXXX becomes the string id number.
```

anon569: Oops.

Issue #1697, opened 2019-05-02
String 6307


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#9128.

```
STR_6307    :Colour scheme: {BLACK}{STRINGID}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [x] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Nubbie @Goddesen @octaroot @telk5093 @Omaranwa @danidoedel @LPSGizmo @BobbyS38 @frenchiveruti @anon569 @JoelTroch

anon569:
Already included in hu-HU

J0anJosep:
ca-ES already done.

galfisk:
Applied to sv-SE

PR #1711, opened 2019-05-12
en-US: Apply #1691, #1697

realConfol:
Applying for issues:
- #1691 
- #1697


data/language/en-US.txt
```diff
@@ -405,7 +405,7 @@ STR_1726    :Land not for sale!
 STR_1727    :Construction rights not for sale!
 STR_1728    :Can't buy construction rights here...
 STR_1729    :Land not owned by park!
-STR_1730    :{RED}Closed - - 
+STR_1730    :{RED}Closed
```
Gymnasiast: This string can be removed altogether, since it's the same as `en-GB`.

PR #1733, opened 2019-07-09
Update sv-SE

AutomCoding:
All strings with ID numbers should now be up to date with en-GB. Some new strings added, a few old ones removed and a lot of updated strings.


data/language/sv-SE.txt
```diff
@@ -212,33 +212,33 @@ STR_0806    :28
...
 STR_0822    :Kan inte komma åt grafikfilen
 STR_0823    :Saknad eller otillgänglig datafil
 STR_0824    :{BLACK}❌
 STR_0825    :Det valda namnet används redan
 STR_0826    :För många namn är definierade
-STR_0827    :Inte tillräckligt med pengar - det behövs {CURRENCY2DP}
+STR_0827    :Inte tillräckligt med pengar – {CURRENCY2DP} behövs
 STR_0828    :{SMALLFONT}{BLACK}Stäng fönster
-STR_0829    :{SMALLFONT}{BLACK}Fönstertitel - Dra för att flytta fönstret
+STR_0829    :{SMALLFONT}{BLACK}Fönstertitel – Dra för att flytta fönstret
```
Gymnasiast: Have you tested whether this character actually works in OpenRCT2? I don't think it does.

(I can wire it to the regular dash, but that requires an update to the code before this is merged.)

data/language/sv-SE.txt
```diff
@@ -212,33 +212,33 @@ STR_0806    :28
...
 STR_0830    :{SMALLFONT}{BLACK}Zooma in
 STR_0831    :{SMALLFONT}{BLACK}Zooma ut
 STR_0832    :{SMALLFONT}{BLACK}Rotera vyn 90° medurs
 STR_0833    :{SMALLFONT}{BLACK}Pausa spelet
-STR_0834    :{SMALLFONT}{BLACK}Inställningar
-STR_0839    :{UINT16} x {UINT16}
-STR_0840    :{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{UINT16} x {UINT16}
+STR_0834    :{SMALLFONT}{BLACK}Spel- och lagringsinställningar
+STR_0839    :{UINT16} × {UINT16}
```
Gymnasiast: This is more correct, but the × character is not in the OpenRCT2 sprite font. I can create a glyph for it, but again, that would need to be added to the OpenRCT2 code before merging this.

data/language/sv-SE.txt
```diff
@@ -2729,81 +2714,81 @@ STR_5217    :Fusk
...
 STR_5256    :Skapa ett nytt tema att göra ändringar på
 STR_5257    :{SMALLFONT}{BLACK}Skapa ett nytt tema baserat på denna
 STR_5258    :{SMALLFONT}{BLACK}Ta bort det använda temat
-STR_5259    :{SMALLFONT}{BLACK}Döp om det använda temat
+STR_5259    :{SMALLFONT}{BLACK}Byt namn på det använda temat
 STR_5260    :Gigantisk skärmdump
 STR_5261    :Filter
-STR_5262    :Wacky Worlds
-STR_5263    :Time Twister
+STR_5262    :Wacky worlds
```
Gymnasiast: These are names, and both are spelt with capitals.

Gymnasiast:
In general, I am very happy with this PR. Especially the changes to compound words - I always thought these should have written without spaces (like in all Germanic languages except English), but since I wasn't a native speaker, I didn't say anything before.

It does look as if you used characters that are not in the sprite font, which will appear as `?`. I have also spotted one mistake (the WW/TT spelling).

I am willing to extend the sprite font in order to support the em-dash and multiplication cross, but since a release is imminent, I'd prefer if you separated those changes from this PR and created a separate one to change these. That way, the 0.2.3 can benefit from the bulk of improved translations.

AutomCoding:
OK, I've reverted my changes to Wacky Worlds and Time Twister now.

When I tested, I saw that N-dash was not supported in banner text and unmade the change in https://github.com/OpenRCT2/Localisation/pull/1733/commits/0d293a571da34a7a9e0f9bd6afc8c31854139699. In such small font as the banner text I don't think it matters which character is used, and only changed because N-dash (*halvfyrkantsstreck*, half-square line) is the "correct" char for horizontal separation in Swedish. (M-dash (*helfyrkantsstreck*, whole-square line) is used for for vertical separation.)

AutomCoding:
As for × (MULTIPLICATION SIGN) and – (EN DASH) in other strings, I doubt many will know the difference from x (LATIN SMALL LETTER X) and - (HYPHEN-MINUS). They are both frequently used in Swedish so I assumed it would not cause any problems.

Should I replace the characters? Either way will do fine for me, although it is recommended to use an asterisk rather than the letter x as multiplication sign.

Gymnasiast:
In general, I think the files should use the correct way of encoding stuff. This means that the sprite font should be able to handle the en-dash and multiplication signs, even if it just wires them to - and x. I'll take care of adding these, you don't need to remove their usages.

AaronVanGeffen:
Were the initial quotes deliberately changed from `“` to `”`? This seems a little odd from a typography perspective.

Gymnasiast:
@AaronVanGeffen They're correct for Swedish. Not every language uses the English style.

PR #1735, opened 2019-07-16
ko-KR: Modify some strings in multiplayer config

telk5093:
Change EOL from LF into CRLF as the same as en-GB


Gymnasiast:
All files should be LF only. If the en-GB file contains CRLF, then that one should be changed.

telk5093:
@Gymnasiast I understand, changed.

PR #1749, opened 2019-09-18
Rewrite some PT-BR strings to avoid truncation

tupaschoal:
Closes OpenRCT2/OpenRCT2#9997


Gymnasiast:
As mentioned on the Localisation channel, it might be better to make the description field larger instead.

This can be kept open in the meantime, though.

tupaschoal:
Ok. The strings didn't lose any information, I just rewrote them to be more concise, but your call :)

PR #1891, opened 2020-06-21
Improve issue and pull request templates

tupaschoal:
Just upgrading the issue templates to the newest format.

Also making the explanation on the PR template a comment, so we  don't need to erase it every time.


Gymnasiast:
One problem is that there are many PRs that change existing strings _and_ introduce new ones.

.github/ISSUE_TEMPLATE/changed_field_template.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Changed Strings
+about: Advertise updates on pre-existing strings that were merged on OpenRCT2
+title: 'String XXXXX'
```
Gymnasiast, line 4:
```suggestion
title: 'String XXXX'
```

Strings have four digits, and will have for some time to come.

Also change this for the other templates.

tupaschoal: Done

PR #1894, opened 2020-06-23
Fix new_and_changed issue template

tupaschoal:
This is the only template that wasn't recognized, you can see that the header formatting is broken at
https://github.com/OpenRCT2/Localisation/blob/master/.github/ISSUE_TEMPLATE/new_and_changed_field_template.md

But is not at this one:
https://github.com/tupaschoal/Localisation/blob/a4e61b39dfe3d8358abb09510294fe06a3ac3a0e/.github/ISSUE_TEMPLATE/new_and_changed_field_template.md

So I guess this will make it work :D


Gymnasiast:
Nope. I fixed your fix.

tupaschoal:
Yeah, the documentation is very poor, they assume you'll use the GH interface for writing the templates.

I'm just glad it works now :)

Issue #1898, opened 2020-07-14
String 5558


tupaschoal:
Updated string on PR OpenRCT2/OpenRCT2#12254.

```
STR_5558    :This setting will only take effect after you have restarted OpenRCT2
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @BobbyS38 @izhangfei @daihakken

tupaschoal:
`pt-BR` translation is good as is already

Nubbie:
sv-SE have already applied

J0anJosep:
ca-ES already applied

Issue #1887, opened 2020-06-21
String 6375


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#11962.

```
STR_6375    :Unknown Ride
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @BobbyS38 @izhangfei @daihakken

galfisk:
Applied to sv-SE

PR #1902, opened 2020-07-23
Update da-DK.txt

LPSGizmo:
Applying for issue(s):
- #1887 
- #1888 
- #1898 

And some missing translations

### DIFF
--- a/data/language/da-DK.txt
+++ b/data/language/da-DK.txt
@@ -1,4 +1,4 @@
-﻿# STR_XXXX part is read and XXXX becomes the string id number.
+# STR_XXXX part is read and XXXX becomes the string id number.
 # Everything after the colon and before the new line will be saved as the string.
 # Use # at the beginning of a line to leave a comment.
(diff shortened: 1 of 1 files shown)


data/language/da-DK.txt
```diff
@@ -1,4 +1,4 @@
-﻿# STR_XXXX part is read and XXXX becomes the string id number.
+# STR_XXXX part is read and XXXX becomes the string id number.
```
Gymnasiast, line 1: Did your editor insert a byte order mark (BOM) here? If so, tell it not to.

LPSGizmo: Sorry my mistake.
i did a copy/paste of the text to edit it offline.
I guess it got lost, when i pasted it back

Gymnasiast: I did some checks and it turns out the you _removed_ the BOM. That is correct, as UTF-8 files should not contain one.

So all good :+1:

PR #1913, opened 2020-08-10
Added *.iml to .gitignore

marcinkunert:
Adding this would make my future translations easier


.gitignore
```diff
@@ -13,4 +13,7 @@ data/language/en-GB.txt
 
 # CLion
 .idea
-cmake-build-debug
\ No newline at end of file
+cmake-build-debug
+
+# IntelliJ IDEA
+*.iml
```
Gymnasiast, line 19: Could you add a newline at the end? We tend to do this throughout where-ever possible.

marcinkunert: Ok, done.

Issue #1942, opened 2020-10-03
Strings 6387-6389


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#13066.

```
STR_6387    :Can't lower element here...
STR_6388    :Can't raise element here...
STR_6389    :No clearance
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [X] en-US
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @BobbyS38 @izhangfei @daihakken @marcinkunert

galfisk:
Applied to sv-SE

PR #1944, opened 2020-10-03
cs-cz apply #1942

octaroot:


Gymnasiast:
Hi, could you use the format `cs-CZ: Apply #1942` next time? (It would save me a bit of time.)

Gymnasiast:
@Oh, another thing. The issue in question has been updated because the last string was wrong. Could you do an update?

Gymnasiast:
@octaroot Oh another thing. The issue in question has been updated because the last string was wrong. Could you do an update?

octaroot:
No update needed, my translation is broad enough to express the updated string

PR #1945, opened 2020-10-03
es-ES: Apply #1942

fernandosutter:
Applying for issue(s):
- #1942

### DIFF
--- a/data/language/es-ES.txt
+++ b/data/language/es-ES.txt
@@ -3649,6 +3649,9 @@ STR_6379    :Se recibieron datos no válidos
 STR_6384    :Nieve
 STR_6385    :Fuertes nevadas
 STR_6386    :Tormenta de nieve
+STR_6387    :No se puede bajar el elemento aquí…
+STR_6388    :No se puede subir el elemento aquí…
+STR_6389    :Sin espacio libre
 
 ##############
 # Escenarios #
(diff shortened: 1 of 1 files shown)


data/language/es-ES.txt
```diff
@@ -3649,6 +3649,9 @@ STR_6379    :Se recibieron datos no válidos
 STR_6384    :Nieve
 STR_6385    :Fuertes nevadas
 STR_6386    :Tormenta de nieve
+STR_6387    :No se puede bajar el elemento aquí…
+STR_6388    :No se puede subir el elemento aquí…
+STR_6389    :Sin espacio libre
```
Gymnasiast, line 3654: This string has been updated. "No clearance" is correct - the error is emitted if the clearance height is not valid at all (below the bottom of the map, or above the top).

data/language/es-ES.txt
```diff
@@ -3649,6 +3649,9 @@ STR_6379    :Se recibieron datos no válidos
 STR_6384    :Nieve
 STR_6385    :Fuertes nevadas
 STR_6386    :Tormenta de nieve
+STR_6387    :No se puede bajar el elemento aquí…
+STR_6388    :No se puede subir el elemento aquí…
+STR_6389    :Sin espacio
```
Gymnasiast, line 3654: Hm, it's not so much that there is no clearance, it's the clearance value itself is invalid - it's a value that cannot exist.

Gymnasiast:
Merged, thanks!

PR #1950, opened 2020-10-07
Keep user list for notifications sorted by language

tupaschoal:
When I tidied up the list, I listed the usernames in the same order as they appear in the readme, which might make things a little easier to maintain/change in the long term (idk)


Gymnasiast:
I don't think it will make any difference, but I have no objections.

PR #1964, opened 2020-10-11
[Issue 13102] Use length signed for all localizations

iking96:
## Description

Related OpenRCT2 PR: #13152
Related OpenRCT2 issue: https://github.com/OpenRCT2/OpenRCT2/issues/13102

Use "LENGTH_SIGNED" tag for altitude ride stat string for all localizations.


Gymnasiast:
We want to modify the `{LENGTH}` tag itself to be a signed 32-bit integer, so the translation files should remain untouched.

This is, by the way, the reason it's usually best to wait with creating a PR against Localisation until the PR at our main repository is merged - you might end up doing work that later turns out to be unnecessary.

PR #1977, opened 2020-10-14
Check action after pull request

marcinkunert:
I've created a GitHub Action to check the translation status. It compares the state before and after a pull request and gives a response as a comment with a table. If some values changed it will be shown as (+ or -) next to the value.

It uses [en-GB.txt](https://github.com/OpenRCT2/OpenRCT2/blob/develop/data/language/en-GB.txt) as a reference.

The table shows:
- how many translations are missing
- how many translations are the same as in the `en-GB.txt`
- how many translations are in the language file, but are not in `en-GB.txt` (and should be probably removed)

To see it in action please visit: https://github.com/marcinkunert/OpenRCT2_Localisation/pull/2

For now it only works for the translations in `key:value` format. It's not working for the scenarios.

This is my first approach to GitHub Actions and Python. Any suggestions are welcome.


Gymnasiast:
This is a nice surprise! :+1: 

It looks very useful to me.

Some small things I'd like to see changed:
1. Could you make it so it only comments on the languages that the PR touches, instead of all of them?
2. "Same as en-GB" implies the string is exactly the same, even though it only means that the string exists in either. This should be renamed.
3. "Not in en-GB" should ignore blank lines and comments (which start with a hash (\#))

marcinkunert:
@Gymnasiast 

1. I've split the table in two. The first shows the changed languages and the second one is hidden and can be expanded.
2. This column counts the same translation as original string for given language. Lets say in both `en-GB` and `de-DE` `STR0005` is `Weather`. That will increase the counter by 1. If the key differ it will not be counted. I've changed the columns to be expandable to explain what it means a bit better.
3. I believe I'm already skipping these lines. (`if not line.startswith('#') and line.find(':') != -1:`)

Here is an example of the new version: https://github.com/marcinkunert/OpenRCT2_Localisation/pull/2#commitcomment-43248874

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
+import filecmp
+
+languages = ['ar-EG', 'ca-ES', 'cs-CZ', 'da-DK', 'de-DE', 'en-US', 'eo-OO', 'es-ES', 'fi-FI', 'fr-FR', 'hu-HU', 'it-IT',
+             'ja-JP', 'ko-KR', 'nb-NO', 'nl-NL', 'pl-PL', 'pt-BR', 'ru-RU', 'sv-SE', 'tr-TR', 'zh-CN', 'zh-TW']
```
tupaschoal, line 4: We should probably read from the filenames on `data/language/*.txt` because then we won't have to remember to update this list when a new language gets added.

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
...
+    with open(filename, encoding="utf8") as f:
+        content = f.readlines()
+    lines = [line.strip() for line in content]
+    translations = {}
+    for line in lines:
+        if not line.startswith('#') and line.find(':') != -1:
+            split = line.split(':')
+            key = split[0].strip()
+            value = ':'.join(split[1:]).strip()
+            translations[key] = value
```
tupaschoal, line 17: So this is currently overwriting the value every time it reads `STR_SCNR`, `STR_PARK` or `STR_DTLS`. In a high-level (and a bit hackish) I think we can support the scenarios by doing:
```python
if line.startswith('#') or len(line.strip()) == 0:
    continue
else if line.startswith('<'):
    # Save scenario_key
else if re.match("STR_\d+", line):
    # Simple strings
else if re.match("STR_(SCNR|PARK|DTLS)", line):
    # Scenario string
    # key = scenario_key + match
```

There are definitely better ways of doing this, such as using `re.compile` for the patterns first, I'm not the keenest in Python, but this is just to try to find a solution for the scenario strings.

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
...
+            elif en_gb[base_string] == translations[base_string]:
+                same_counters[lang] += 1
+                if print_info:
+                    print('[{}] Same translation: {}'.format(lang, base_string))
+
+        for translation in translations:
+            if translation not in en_gb:
+                not_in_base_counters[lang] += 1
+                if print_info:
+                    print('[{}] Unnecessary translation: {}'.format(lang, translation))
```
tupaschoal, line 45: I'm wondering what would be better:
 - Doing like what is already being done
 - Saving all missing/same/unnecessary strings into the object using list comprehension, paying the price of iterating three times, and then printing these lists at the end
 - Saving all missing/same/unnecessary strings into the object at each iteration and then printing these lists at the end

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
...
+    same_counters = dict.fromkeys(languages, 0)
+    not_in_base_counters = dict.fromkeys(languages, 0)
+
+    for lang in languages:
+        translations = file_to_dict(dir_with_translations + '/' + lang + '.txt')
+        for base_string in en_gb:
+            if base_string not in translations:
+                missing_counters[lang] += 1
+                if print_info:
+                    print('[{}] Missing translation: {}'.format(lang, base_string))
```
tupaschoal, line 35: I guess using f-strings look clearer
```suggestion
                    print(f'[{lang}] Missing translation: {base_string}')
```

This would apply to all `format` calls in the patch

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
...
+        missing = '**{0}**'.format(missing)
+    return missing
+
+
+def run():
+    master_branch = count_translations('master/data/language', False)
+    pr = count_translations('pr/data/language', True)
+
+    missing = '<p><details><summary>Missing</summary>The translation is not added to translation file. (e.g. STR_9999 is in `en-GB` but is not available in given language)</details></p>'
+    same = '<p><details><summary>Same</summary>The translation and source string is exactly the same.  (e.g. in STR_9999 is `Umbrella` in both `en-GB` and given language)</details></p>'
```
tupaschoal, line 65:
```suggestion
    same = '<p><details><summary>Same</summary>The translation and source string is exactly the same.  (e.g. STR_9999 is `Umbrella` in both `en-GB` and given language)</details></p>'
```

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,103 @@
...
+    not_needed = '<p><details><summary>Not needed</summary>The translation file contains entries that are not in `en-GB` and should be removed (e.g. STR_9999 exits in given language but is not in `en-GB`)</details></p>'
```
tupaschoal, line 66: These lines are very similar, I would probably extract it to a function that does something like:

```python
def get_table_header(title, details):
    return f'<p><details><summary>{title}</summary>{details}</details></p>'

missing = get_table_header("Missing", "The translation is not added to translation file. (e.g. STR_9999 is in `en-GB` but is not available in given language)")
same = get_table_header("Same", "The translation and source string is exactly the same.  (e.g. in STR_9999 is `Umbrella` in both `en-GB` and given language)")
not_needed = get_table_header("Not needed", "The translation file contains entries that are not in `en-GB` and should be removed (e.g. STR_9999 exits in given language but is not in `en-GB`)")
```

On a side note, I like that the details are collapsed by default. It would be awesome if they also appeared on the tooltip of the summary.

tupaschoal (commented):
Nice surprise indeed!

> This is my first approach to GitHub Actions and Python

Not sure if you meant first approach with the combination or if you're also new to Python. In any case I'd recommend running `pylint` on the script, it will at least recommend some docstrings for better documentation

.github/workflows/translation_check.py
marcinkunert: I've suspected something like this exists in Python, but couldn't find it.

.github/workflows/translation_check.py
marcinkunert: Good idea. I'll check in both `master` and PR branch.

marcinkunert:
>Not sure if you meant first approach with the combination or if you're also new to Python. In any case I'd recommend running pylint on the script, it will at least recommend some docstrings for better documentation

Both. This is my first Python script (I've started with looking up how to print Hello World) and my first GitHub Actions. I'm using JetBrains PyCharm and it helps with formatting, but I'll look into pylint.

.github/workflows/translation_check.py
marcinkunert: GFM (GitHub Flavored Markdown Spec) doesn't seem to support tooltips

https://stackoverflow.com/a/49333886/1756721

.github/workflows/translation_check.py
marcinkunert: I've tested one version which saves the missing/same/unnecessary to be later shown as a foldable block. Unfortunately there is a limit in the github comment size and it couldn't fit so I've reverted. 
I've also looked a bit into publishing action artifacts, but it is only available as a zip download and using the build-in option for searching in logs seems a lot more convenient.
For now I see no benefit of first saving them, to be later displayed.

marcinkunert:
I've improved the script.

- added support for scenarios and objects
- added automatic detection of translations
- changed string formatting to f-string
- added docs
- fixed issues suggested by pylint

.github/workflows/translation_check.py
marcinkunert: I've added support to scenarios and objects(?). Done similar as suggested.

.github/workflows/translation_check.py
marcinkunert: Extracting similar strings is a good idea. Done.

.github/workflows/translation_check.py
marcinkunert: Done.

Gymnasiast:
Could you move "Not needed" to the middle, and amend the note for "Same" to emphasise that this might mean the string is untranslated, but also that it might just happen to be the same. (We unfortunately still have a lot of "technical strings" that should not be in the language files, but are there for technical reasons.)

Otherwise, looking good AFAIC.

marcinkunert:
@Gymnasiast done. https://github.com/marcinkunert/OpenRCT2_Localisation/pull/2#commitcomment-43280999

.github/workflows/.pylintrc
```diff
@@ -0,0 +1,2 @@
+[FORMAT]
+max-line-length=120
```
Gymnasiast, line 2: Please end your file with a newline. (We do this for all of our files.)

Gymnasiast (approved):

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,177 @@
...
+It checks the difference in translation count between master and PR branch.
+The results is put to result.md to be later published as a comment.
+"""
+
+import filecmp
+import os
+import re
+
+languages = []
+
```
tupaschoal, line 12:
```suggestion
MASTER_LANG_DIR = "master/data/language"
PR_LANG_DIR = "master/data/language"

```
Define these as global constants and reuse elsewhere in the code, so it becomes easier to change later if needed.

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,177 @@
...
+            other_table += row
+
+    result += prepare_spoiler('Other translations', other_table)
+
+    text_file = open("result.md", "w")
+    text_file.write(result)
+    text_file.close()
+
+
+run()
```
tupaschoal, line 177:
```suggestion
if __name__ == "__main__":
    run()
```

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,177 @@
...
+        same = format_result(master_branch['same'][lang], pull_request['same'][lang])
+
+        row = f'|`{lang}`|{missing}|{not_needed}|{same}|\n'
+
+        if lang in languages_changed:
+            result += row
+        else:
+            other_table += row
+
+    result += prepare_spoiler('Other translations', other_table)
```
tupaschoal, line 170: I feel like this should be extracted to a `get_translation_report` or similar

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+    if not os.path.exists(filename):
+        return {}
+
+    with open(filename, encoding="utf8") as file:
+        content = file.readlines()
+    lines = [line.strip() for line in content]
+    translations = {}
+
+    special_keys = ['STR_SCNR', 'STR_PARK', 'STR_DTLS', 'STR_NAME']
```
tupaschoal, line 45: Define this along with the globals at the top (use UPPERCASE for globals)

tupaschoal (commented):
This looks nice, almost ready to go, thanks for using the pylint and improving it.

What I suggested shouldn't change anything functionally, just become a little bit more modular.

.github/workflows/translation_check.py
```diff
@@ -0,0 +1,174 @@
...
+    same = prepare_spoiler('Same', 'The translation and source string is exactly the same.  '
+                                   '(e.g. STR_9999 is `Umbrella` in both `en-GB` and given language). '
+                                   'This may be desired in some cases (e.g. `April` is the same in English and German)')
+    table_header = "| |" + missing + " | " + not_needed + " |" + same + "|\n"
+    table_header += "|---|---|---|---|\n"
+    languages_changed = []
+    for lang in languages:
+        filename = lang + '.txt'
+        file_master = MASTER_LANG_DIR + '/' + filename
+        file_pr = MASTER_LANG_DIR + '/' + filename
```
tupaschoal, line 145:
```suggestion
        file_master = os.path.join(MASTER_LANG_DIR, filename)
        file_pr = os.path.join(MASTER_LANG_DIR, filename)
```

Use `os.path.join` to make sure the proper file separator is picked (depending on the OS)

tupaschoal (commented):
Looks good. I think there are a few things here and there that can be improved but are not major, so we can pick it up later, I want to see the script up and running.

I've left just one minor comment to wrap up

tupaschoal:
Ah, one github flow tip for the future, when you do solve the issues raised by the reviewers, make sure you mark the conversation as resolved, so we know you actually tended to that comment :)

marcinkunert:
@tupaschoal thanks for the tip, all done

tupaschoal (approved):
Looks ok now, let's merge and see how it goes :rocket:

Thanks!

Issue #1979, opened 2020-10-16
Strings 6394–6395 (+ various changed strings)


AaronVanGeffen:
Updated strings on PR OpenRCT2/OpenRCT2#13186.

Rephrased strings:

```diff
-STR_2753    :Mowed grass
+STR_2753    :Mow grass
-STR_2767    :Freeze weather
+STR_2767    :Disable weather changes
-STR_5725    :{BLACK}Force weather:
+STR_5725    :{BLACK}Change weather to:
-STR_5726    :{SMALLFONT}{BLACK}Sets the current weather in park
+STR_5726    :{SMALLFONT}{BLACK}Sets the current weather in the park
-STR_5871    :Plants don't age
+STR_5871    :Disable plants withering
```

Capitalisation changes:

```diff
-STR_2765    :Large Tram
+STR_2765    :Large tram
-STR_2769    :Open Park
+STR_2769    :Open park
-STR_2770    :Close Park
+STR_2770    :Close park
-STR_6331    :Create Ducks
+STR_6331    :Create ducks
-STR_6332    :Remove Ducks
+STR_6332    :Remove ducks
```

New strings:
```
STR_6394    :Objective
STR_6395    :Maintenance
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @JasperGray @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @izhangfei @daihakken @marcinkunert

frenchiveruti:
What's "Large Tram" or what's the context, because in spanish someone before me translated STR_2765 as "Visitors group"

tupaschoal:
Clicking it spawns a looong queue of peeps

[...]

PR #1984, opened 2020-10-17
es-ES: Apply #1979

frenchiveruti:
Applying for issue(s):
- #1979


frenchiveruti:
I removed the commit that was useless.

PR #1987, opened 2020-10-17
pl-PL apply #1979

marcinkunert:
Applying for issue #1979
I'll keep it as a draft in order to test the translation check action

### DIFF
--- a/data/language/pl-PL.txt
+++ b/data/language/pl-PL.txt
@@ -8,7 +8,6 @@ STR_0003    :Stojąca kolejka górska
 STR_0004    :Podwieszana kolejka górska
 STR_0005    :Odwrócona kolejka górska
 STR_0006    :Dziecięca kolejka górska
-STR_0007    :Miniaturowa kolejka
 STR_0008    :Kolejka jednotorowa
 STR_0009    :Podwieszana mini kolejka
 STR_0010    :Łódki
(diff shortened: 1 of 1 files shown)


tupaschoal:
(seems like something is wrong in the script)

marcinkunert:
The script is fine, but there seems to be a limitation to the commit action so that it only works if the PR author has edit rights for the repo (https://github.com/actions/first-interaction/issues/10). I'm currently looking for a fix for that.

tupaschoal:
I think you need to use the GH actions secret

marcinkunert:
That seems to be the case. I'll test that out on my fork. Afterwards I would like to ask you to create a token with write rights to this repo and setup it as a secret.

tupaschoal:
I guess you can use the default built-in token every repo has, `secrets.GITHUB_TOKEN`

marcinkunert:
This one gets used by default but lacks write access to PR's which is apparently needed.

tupaschoal:
I think it'd work because the PR runs on upstream not the fork. I guess that column is if you were trying to run it in a pr in your own repo (?), Not sure though

marcinkunert:
Hopefully fixed in #1989

marcinkunert:
Closing this PR. I'll create a new one in order to check if the translation check does work now as intended.

PR #1988, opened 2020-10-17
de-DE: Apply #1979

Wuzzy2:
Applying for issue(s):
- #1979


Wuzzy2:
So how does this new translation check feature work? Can you tell me if de-DE has 100% coverage?

marcinkunert:
@Wuzzy2 its broken at the moment, I'm working on fixing it

Gymnasiast:
It should automatically tell you on the pull request, but it seems the rights are not set quite right.

tupaschoal:
It should tell you. We just merged it and it still needs polishing, but if you expand the "run checks" tab of its run and search for your language code, you can get some insight. Seems like some stuff is missing indeed.

PR #1989, opened 2020-10-17
Fix translation check action 

marcinkunert:
Changed translation check action trigger in order to work around a lack of write access when executed on fork. Hopefully this does the trick.

### DIFF
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -1,8 +1,6 @@
 name: Translation Check
 
-on:
-  pull_request:
-    branches: [ master ]
+on: pull_request_target
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


Gymnasiast (approved):

marcinkunert:
> Check has not run at all

I'm guessing that's because this particular PR touches the workflow which should be triggered. Either way if the previous trigger is broken than this PR has a chance to fix it.

tupaschoal:
Maybe touch some language files in a separate commit so we can test it on the pr, then we drop that commit later before merging

marcinkunert:
ASASIK GitHub will trigger the workflow which is on the master branch. There is a working example on my fork: https://github.com/marcinkunert/OpenRCT2_Localisation/pull/7 
Come on guys it's broken anyways atm, why not simply approve the pr and check what will happen?

tupaschoal:
I understand it can be frustrating to have the check broken. We just want to get it right with fewer back and forth steps, in case it's not the final solution.

marcinkunert:
If you want to help out then creating a PR to this repo: https://github.com/marcinkunert/OpenRCT2_Localisation should be a good way of testing if this actually works.

tupaschoal:
I've read a bit about the `pull_request_target` and I think the change makes sense. As far as I'm concerned, it doesn't run in this pr because it's a protection mechanism from GitHub

tupaschoal (approved):

PR #1990, opened 2020-10-17
pl-PL: Apply #1979

marcinkunert:
Applying for issue: #1979

Hopefully translation check works as intended now.


marcinkunert:
Yay it worked 🎉

Two quirks:

- One failed test - GitHub probably saved the result from previous check on that commit. 
- Changes on languages other than `pl-PL` - my branch wasn't rebased completely. Not sure what I can do to avoid this though other than keeping the repo rebased

marcinkunert:
Looks a lot better after rebasing

Gymnasiast:
Most translators are not very good with Git. We cannot expect them to rebase. The results really have to be correct even if the branch is not rebased.

marcinkunert:
I agree. Opened an issue for that #1991

PR #1992, opened 2020-10-18
Check against fork master instead of upstream master

### LINKED ISSUE #1991: Improve check action to work with not rebased branches
Currently the check compares current master vs and the PR branch. This doesn't work properly if the PR branch is not rebased. Example can be found here: https://github.com/OpenRCT2/Localisation/pull/1990

This should be improved and the check should be done probably between PR newest commit and PR branch start (?). I'm not sure though at the moment. To be investigated.

marcinkunert:
The translation check will be now done against the fork master branch and not the upstream master. This should prevent showing changes in case a not fully rebased branch is send to a PR. Resolves #1991

image

### DIFF
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -21,11 +21,21 @@ jobs:
           ref: develop
           path: OpenRCT2
 
-      - name: Get translations master (for comparsion)
+      - name: Checkout translations before PR changes (for comparsion)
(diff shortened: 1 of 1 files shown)


Gymnasiast:
This will still fail if the translator commits to their own `master` branch. I'd like a fix for that as well, but in the meantime we will at least have statistics for >90% of PRs.

PR #2001, opened 2020-10-22
Remove unused strings from translations

tupaschoal:
- Most of those had the same set of strings that aren't in `en-GB` anymore. 
- There were a couple fixes to some scenario names (some were using the alias, but I guess it's better not to, to make the script result sound).
- There was a lot of extra content on `nl-NL` not found elsewhere, which I'm sure @Gymnasiast can explain (there was some on `ko-KR` too.


Gymnasiast:
> There was a lot of extra content on nl-NL not found elsewhere, which I'm sure @Gymnasiast can explain (there was some on ko-KR too.

Content which should have stayed there. It translates some common custom objects that had no Dutch translation. I don't think it was a particularly good idea to remove it before asking.

I merged your PR before I saw that, because I would not have done that otherwise. I have now removed the commit in question. Please open a new PR that does not touch the object descriptions.

tupaschoal:
Argh, but I even commented that :(

tupaschoal:
> Much of the content (apart from the official custom content, which has since been put into JSON objects) should have stayed there.

@Gymnasiast can  you point out what should stay?
Is there anything that has since been moved to objects?
Should this be in `en-GB` also?

Gymnasiast:
- `Panda World` (an official DLC scenario) - could be in `en-GB` as well, as that makes translators more likely to translate it.
- anything marked as OCC (official custom content) can go (this means Panda stuff and pink water)
- objects present [here])https://github.com/OpenRCT2/objects/tree/master/objects/official) may be removed

That leaves some custom objects. I'd prefer making these into JSON objects and make sure that they can be autodownloaded by the object downloader, but we have no mechanism for that yet - only for objects that are included in OpenRCT2 by default. For now though, I think it's best to leave their translations as they are.

PR #2002, opened 2020-10-22
Improve translation_check.py usability

tupaschoal:
Ping @marcinkunert 

This adds some quality of life improvements to `translation_check`, along with some minor refactors:
- Extracts hardcoded path to `en-GB.txt`
- Sorts the language list, so that the output of the check and the comments always list languages alphabetically
- Adds an argparser, so it can be run locally by doing, for example:
`python3 translation_check.py -r data/language/en-GB.txt -m data/language -b data/language`
The default values of the arguments are the hardcoded ones, so that we don't have to use the switches in the workflow. Also note that having this provides an easy way to run the script locally, which makes it a lot faster to validate changes in the script

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,3 +78,21 @@ Anyone can still contribute to the languages.
 ### Questions & Support
 
 If you have any question or issues, you can always ask us on [Discord](https://discord.gg/sxnrvX9).
+
+### Checking Translation Integrity
+
+We have an action that automatically runs on every Pull Request and reports, for every supported language:
+
+- If there are any strings left to translate
+- If there are any strings that do not need translation
+- If there are any strings that might not have been translated, as they are the same
+
+The script is on the repository and can be run locally if one wants to anticipate the results, by doing:
+
+```
+python .github\workflows\translation_check.py -r <path to en-GB.txt> -m <path to data/language folder on master branch> -b <path to data/language folder on your branch>
+```
+
+Note that `-b` and `-m` can use the same argument if you don't want differential analysis.
+
+You can use the `--help` switch to read more about what each switch does.
--- a/.github/workflows/translation_check.py
+++ b/.github/workflows/translation_check.py
@@ -4,12 +4,14 @@ It checks the difference in translation count between master and PR branch.
 The results is put to result.md to be later published as a comment.
 """
 
+import argparse
 import filecmp
 import os
 import re
 
 MASTER_LANG_DIR = "master/data/language"
 PR_LANG_DIR = "pr/data/language"
+OPENRCT2_EN_GB_FILE = "OpenRCT2/data/language/en-GB.txt"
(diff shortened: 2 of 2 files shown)


marcinkunert:
Looks good but please also add documentation how to use it in `README.md`

tupaschoal:
I wonder if it's really needed because the `argparse` already builds the help for us:
```
python translation_check.py -h
usage: translation_check.py [-h] [--reference-lang-file REFERENCE_LANG_FILE]
                            [--master-dir MASTER_DIR]
                            [--branch-dir BRANCH_DIR]

Check translation integrity

optional arguments:
  -h, --help            show this help message and exit
  --reference-lang-file REFERENCE_LANG_FILE, -r REFERENCE_LANG_FILE
                        a reference language file, typically en-GB.txt
                        (default: OpenRCT2/data/language/en-GB.txt)
  --master-dir MASTER_DIR, -m MASTER_DIR
                        the language directory on the master branch (default:
                        master/data/language)
  --branch-dir BRANCH_DIR, -b BRANCH_DIR
                        the language directory on the pull request branch
                        (default: pr/data/language)

```

marcinkunert:
I'm pretty sure that a information that the script is available and single example would allow translators to use it. How otherwise should they know that it even exists? Some of them find using git troublesome so it might implicate that using executing a script with parameters might be also a small challenge.

tupaschoal:
Done

README.md
```diff
@@ -78,3 +78,19 @@ Anyone can still contribute to the languages.
...
+We have an action that automatically runs on every Pull Request and reports, for every supported language:
+
+- If there are any strings left to translate
+- If there are any strings that do not need translation
+- If there are any strings that might not have been translated, as they are the same
+
+The script is on the repository and can be run locally if one wants to anticipate the results, by doing:
+
+```
+python .github\workflows\translation_check.py -r <path to en-GB.txt> -m <path to data/language folder> -b <path to data/language folder>
```
marcinkunert, line 93: -m and -b have the same description. -m could be described as "path to data/language folder on master branch" and -b "path to data/language folder on your branch"  or so.

It could be also worth adding a note that the translator is expected to checkout the repository twice to avoid confusion

tupaschoal: I guess for checking whether something is missing or not `-b` and `-m` can indeed receive the same folder, no? It will not report the difference caused by the PR, but it will still provide most of the other info

(I agree on clarifying what each switch does)

marcinkunert: Yeah, that's true. I've never considered that. In this case the user is more interested in the stdout than the result file. I guess adding a language filter would be very useful in order to avoid printing all the other languages.

tupaschoal: I've improved the explanation.

>  I guess adding a language filter would be very useful in order to avoid printing all the other languages.

That's true, I thought about that yesterday, but I guess I want to see whether people really will start using it or not before providing it (not that it's any difficult though)

PR #2004, opened 2020-10-23
ko-KR: Apply #2005 and remove unused strings

telk5093:
Remove unused strings and Add missings to make the translation check numbers zero that added recently.
I've already seen https://github.com/OpenRCT2/Localisation/pull/2001, but please let me know if there is any problem :)


Gymnasiast:
I see you synchronised your file with en-GB. However, there was some old data in en-GB, and I just cleaned it up. You can see what I changed in https://github.com/OpenRCT2/Localisation/issues/2005.

telk5093:
@Gymnasiast Checked, thanks for notice :)

Issue #1998, opened 2020-10-21
Improve translation check action


marcinkunert:
Few things came up after seeing the script in action. Short intro how it works:

- if the pull reques

tupaschoal:
This looks like a great enhancement and I think it touches on some limitations that @Gymnasiast had raised before.

marcinkunert:
PR from master seems to work now (example here #2010). Only thing to do is to not post a comment if there are no changes or the changes do not affect the translation files at all.

PR #2007, opened 2020-10-24
Correctly handle pull requests from master branch

marcinkunert:
This should fix (or ad least improve) the situation in which a pull request is created from the fork master branch as described here #1998

### DIFF
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -30,12 +30,34 @@ jobs:
           fetch-depth: 0
 
       - name: Checkout commit before branch changes
+        id: merge-base-commit
         run: |
+          echo "head sha: ${{github.event.pull_request.head.sha}}"
           cd master
-          git fetch origin master:refs/remotes/origin/master
-          git merge-base refs/remotes/origin/master HEAD
-          git checkout $(git merge-base refs/remotes/origin/master HEAD)
+          git remote add upstream https://github.com/${{github.repository}}
+          git fetch upstream
+          merge_base=$(git merge-base refs/remotes/upstream/master HEAD)
+          echo "merge base sha: ${merge_base}"
+          git checkout $merge_base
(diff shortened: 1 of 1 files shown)


marcinkunert:
I've also changed so that the script will be run from upstream master, not the fork. This should fix situations like this fail: #2006

.github/workflows/translation_check.yml
```diff
@@ -30,20 +30,48 @@ jobs:
...
 
+      - name: Checkout upstream master (in order to use the newest version of the script)
+        uses: actions/checkout@v2
+        with:
+          path: upstream_master
+          ref: master
+
       - name: Run checks
-        run: python master/.github/workflows/translation_check.py
+        run: python upstram_master/.github/workflows/translation_check.py
```
tupaschoal, line 74: There is a typo here "upstram"

.github/workflows/translation_check.yml
```diff
@@ -30,20 +30,48 @@ jobs:
...
 
+      - name: Print debug stuff
+        run: |
+          echo ${{steps.merge-base-commit.outputs.merge_base}}
+          echo ${{github.event.pull_request.head.sha}}
+          echo ${{steps.merge-base-commit.outputs.merge_base == github.event.pull_request.head.sha}}
+
+      - name: Remove master
+        if: ${{ steps.merge-base-commit.outputs.merge_base == github.event.pull_request.head.sha }}
+        run: rm -rf master
```
tupaschoal, line 53: Not sure we should force it

.github/workflows/translation_check.yml
marcinkunert: That's true :( Fixed in next PR

.github/workflows/translation_check.yml
marcinkunert: This doesn't matter as it only will be deleted in some cases and the directory will be created again in the next step

PR #2017, opened 2020-10-26
pt-BR: Translated some missing strings

tupaschoal:
Using the new GH action, I found out some strings that were never translated in `pt-BR`


data/language/pt-BR.txt
```diff
@@ -3547,18 +3547,18 @@ STR_6290    :Protocolo base
 STR_6291    :Comandos
 STR_6292    :Mapa
 STR_6293    :B
-STR_6294    :KiB
-STR_6295    :MiB
-STR_6296    :GiB
-STR_6297    :TiB
+STR_6294    :kB
```
Gymnasiast, line 3550: This is not correct. 1 kB = 1000 bytes. 1 KiB = 1024 bytes.

tupaschoal: Ah, I see. Thanks!

PR #2031, opened 2020-11-06
Fix #2011: STR_5151 and STR_5152 are not technical strings

### LINKED ISSUE #2011: Translation script should not treat STR_5151 and STR_5152 as technical strings
These strings are the thousands and decimal point separators and differ depending on locale

https://github.com/OpenRCT2/OpenRCT2/blob/9dcbf6d4e62bd13cf167b63bab69bfd26c20e10c/src/openrct2/localisation/StringIds.h#L2643-L2644

According to this: https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

In German a big number would be written like this: 4 294 967.295,000
In English: 4,294,967,295.00
In Italian: 4.294.967.295,000
In Danish/Finnish/French: 4 294 967 295,000  

There is also a large article on that on wikipedia: https://en.wikipedia.org/wiki/Decimal_separator

Gymnasiast:


tupaschoal (approved):
Ideally we would keep the list tidy, but I understand if you don't want to do it because it's boring

Gymnasiast:
It was mostly to avoid making the diff too big.

I can make the list tidy if you like.

PR #2033, opened 2020-11-07
Remove unused strings from translations

tupaschoal:
- Most of those had the same set of strings that aren't in en-GB anymore.
- There were a couple fixes to some scenario names (some were using the alias, but I guess it's better not to, to make the script result sound).
- There was a lot of extra content on nl-NL not found elsewhere, which I'm sure @Gymnasiast can explain (there was some on ko-KR too.


data/language/nl-NL.txt
```diff
@@ -4509,15 +4421,6 @@ STR_NAME    :Dak
 [TTPIRF08]
 STR_NAME    :Dak
 
-[TTPRF09]
```
Gymnasiast, line 4512: Restore this/

data/language/nl-NL.txt
```diff
@@ -4413,68 +4387,6 @@ STR_SCNR    :Panda World
 STR_PARK    :Panda World
 STR_DTLS    :Voeg meer attracties toe aan dit park met pandathema, en trek meer bezoekers aan.
 
-##################################
```
Gymnasiast, line 4416: Restore this.

data/language/nl-NL.txt
tupaschoal: This is not in the `en-GB` file. Should it be?

data/language/nl-NL.txt
tupaschoal: Do you mean the whole rides/attractions block? Are these not tracked on objects?

Gymnasiast: The whole block, yes.

These objects are not tracked on `objects` - only official ones are. The objects in this file are unofficial objects that are common enough to warrant a translation.

data/language/nl-NL.txt
Gymnasiast: We could consider putting them in `en-GB`. But in the longer run, I want to create objects for these. The problem is that currently, our JSON objects are either supplied with the game, or not available at all. We should really have a system to download JSON objects.

tupaschoal:
I've restored the blocks in a separate commit, for ease of review. Do squash when merging

PR #2056, opened 2020-11-28
eo-OO: Some corrections and object TLs

JasperGray:
- Use the term gokart directly instead of vetkurantaj aŭtetoj (racing minicars)
- Duobla-ferdeka (double-decker) instead of duoble-ferdeka (doubly decked)
- Some objects have inversigitaj (inverted) elements, not renversitaj (overturned) ones
- Pair can be translated more precisely as paro, not duopo (ordered pair)
- Translated some similar object descriptions having to do with riders sitting in pairs of seats,
  to give them consistent translations


Gymnasiast:
Are you able to rebase, or would you like me to do that?

JasperGray:
I'll rebase it on my end, thanks

PR #2063, opened 2020-11-29
eo-OO: Translate objects starting with letter T, P & O

tellovishous:
Plus several other changes


Gymnasiast:
The rebase didn't go quite correct. There should be no merge commit.

Issue #2035, opened 2020-11-08
String 6399


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#13394.

```
STR_6399    :OpenRCT2 needs files from the original RollerCoaster Tycoon 2 in order to work. Please set the “game_path” variable in config.ini to the directory where you installed RollerCoaster Tycoon 2, then restart OpenRCT2.
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @JasperGray @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @izhangfei @daihakken @marcinkunert

Wuzzy2:
de-DE is done.

galfisk:
Applied to sv-SE

Issue #2027, opened 2020-11-01
String 6398


telk5093:
New strings from PR OpenRCT2/OpenRCT2#13256.

```
STR_6398    :File contains unsupported ride types. Please update to a newer version of OpenRCT2.
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @JasperGray @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @izhangfei @daihakken @marcinkunert

Wuzzy2:
de-DE is done.

galfisk:
Applied to sv-SE

PR #2068, opened 2020-12-04
de-DE: STR_6398, STR_6399

Wuzzy2:
Applying for issue(s):
- #2035
-  #2027


Wuzzy2:
So the checker claims 20 strings are missing. How do I know which 20 strings are missing? Is there a FOSS script that I can run?

Wuzzy2:
OK, I found the checker script now. I have added a translation for an old missing string (`STR_6123`). But the other 19 "missing" strings seem to be bogus reports by the checker script (see #2069).

PR #2100, opened 2020-12-20
eo-OO: Nombro da, not nombro de

JasperGray:


JasperGray:
nvm just double-checked. does not apply in this case
https://vortaro.net/#nombro_kd

PR #2102, opened 2021-01-02
ca-ES: Fixed some build checks

gerardnll:


Gymnasiast:
Thanks for the tooltip fix, I have now also applied that to the other languages! :+1:

Issue #2121, opened 2021-01-19
Replace typewriter quotes


Gymnasiast:
The straight quotation marks - `'` and `"` - were introduced by typewriters to save space, which carried over to ASCII.

Since the late 90s, there is little need to use them, as proper quotations have been introduced since. For example, this means you should use  “ ” and ‘ ’ for quotation marks (at least in English, other languages can have different quotation marks such as « ») and ’ as an apostrophe.

More info can be found here: https://en.wikipedia.org/wiki/Quotation_mark

The typewriter quotes should be replaced everywhere except in technical strings such as `<Mel's World>`. And if present, the backtick (\`) should be replaced with ‘.

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @JasperGray @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @Nubbie @izhangfei @daihakken @marcinkunert

Wuzzy2:
I also like to add: The ASCII apostrophe (U+0027) can be replaced with U+2019: https://en.wikipedia.org/wiki/%E2%80%99

Gymnasiast:
> I also like to add: The ASCII apostrophe (U+0027) can be replaced with U+2019: https://en.wikipedia.org/wiki/%E2%80%99

I already said that in the opening post:

> For example, this means you should use “ ” and ‘ ’ for quotation marks ... **and ’ as an apostrophe**.

galfisk:
Applied to sv-SE

Gymnasiast:
@galfisk Also to the objects translation?

galfisk:
@Gymnasiast to everything in the sv-SE file apart from the technical strings, by search-and-replace.

[...]

PR #2123, opened 2021-01-21
de-DE: Replace ASCII apostrophes and backticks

Wuzzy2:
Applying for issue(s):
- #2121


Gymnasiast:
This PR also needs to modify https://github.com/OpenRCT2/Localisation/blob/master/objects/de-DE.json .

Wuzzy2:
OK, done. It was only 3 strings.

Gymnasiast:
Good, thanks!

PR #2125, opened 2021-01-22
eo-OO: Apply #2121

rzhao271:


objects/eo-OO.json
```diff
@@ -9141,43 +9141,43 @@
   },
   "rct2.ww.rwdaub02": {
     "reference-name": "Wattle & Daub Roof",
-    "name": "Tegmento konstruita per 'Wattle k Daub'"
+    "name": "Tegmento konstruita per ‘Wattle k Daub’"
```
Gymnasiast, line 9144: Is there no better way to translate this? Wattle and daub is a construction technique that has been around for 6000 years. It almost looks like you interpreted it as a company name.

See here: https://en.wikipedia.org/wiki/Wattle_and_daub

rzhao271: Yeah, I couldn't find a way to easily translate the words wattle or daub at that time, and I'm still not sure what the best approach is here.
I noticed some Romance languages say torchis, though I don't think that word is in Esperanto either, I'll have to re-check later.
Either I esperantize torchis and then users have to search it up and hopefully hit a Romance language dictionary, or I use the preposition "el" to denote that "Wattle k Daub" is some material, which I should've done anyway.

Gymnasiast: Yeah, but why use English in the first place? There is nothing that tied them to an Anglophone region.

Also, the building materials (fine branches and loam/earth) are incredibly common and old. There *have* to be words for those in Esperanto.

rzhao271: The thought of just translating over the building materials rather than the exact construction technique escaped me for some reason, and there's no good reason why I kept it in English. Will re-translate those lines later today!

PR #2126, opened 2021-01-22
eo-OO: Change maintainer

rzhao271:
I would like to maintain the eo-OO localization with this account now, thanks.


Gymnasiast:
You will also need to modify the three issue templates.

PR #2128, opened 2021-01-25
pt-BR: Apply #2121

tupaschoal:
Applying for issue(s):
- #2121

### DIFF
--- a/data/language/pt-BR.txt
+++ b/data/language/pt-BR.txt
@@ -3347,10 +3347,10 @@ STR_6068    :Servidor desligado.
 STR_6069    :{STRING} foi chutado do servidor por {STRING}.
 STR_6070    :{STRING} foi definido para grupo ‘{STRING}’ por {STRING}.
 STR_6071    :{STRING} criou novo grupo de jogadores ‘{STRING}’.
-STR_6072    :{STRING} deletou grupo de jogadores ‘{String}’.
-STR_6073    :{STRING} editou permissões para grupo de jogadores ‘{String}’.
-STR_6074    :{STRING} mudou nome do grupo de jogadores de ‘{String}’ para ‘{String}’.
-STR_6075    :{STRING} mudou grupo de jogadores padrão para ‘{String}’.
+STR_6072    :{STRING} deletou grupo de jogadores ‘{STRING}’.
+STR_6073    :{STRING} editou permissões para grupo de jogadores ‘{STRING}’.
+STR_6074    :{STRING} mudou nome do grupo de jogadores de ‘{STRING}’ para ‘{STRING}’.
+STR_6075    :{STRING} mudou grupo de jogadores padrão para ‘{STRING}’.
 STR_6076    :{STRING} usou/alternou trapaça ‘{STRING}’.
 STR_6077    :Adicionar Dinheiro
 STR_6078    :{STRING} criou atração ‘{STRING}’.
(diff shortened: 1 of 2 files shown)


objects/pt-BR.json
```diff
@@ -1,19 +1,19 @@
 {
   "rct2.glthent": {
     "reference-name": "'Goliath' Sign",
-    "name": "Placa 'Golias'"
+    "name": "Placa ‘Golias’"
```
Gymnasiast, line 4: "Goliath" refers to a roller coaster with that name, not to the biblical giant.

Gymnasiast:
There are still some `'` quotes in the scenario translations, for example on line 4458 (Rock 'n' Roll - Rock 'n' Roll).

BTW, strictly speaking, `Rock 'n' Roll` should be rendered as `Rock ’n’ Roll` - two apostrophes, since we're omitting letters.

objects/pt-BR.json
tupaschoal: Thanks!

data/language/pt-BR.txt
```diff
@@ -3345,34 +3345,34 @@ STR_6066    :Registra todas as ações do usuário em arquivos no diretório do
 STR_6067    :Servidor iniciado.
 STR_6068    :Servidor desligado.
 STR_6069    :{STRING} foi chutado do servidor por {STRING}.
-STR_6070    :{STRING} foi definido para grupo'{STRING}' por {STRING}.
```
tupaschoal, line 3348: Seems like the two {STRING} at the end of those lines should be surrounded by simple quotes. They aren't in the original too, so I'm opening for discussion.

data/language/pt-BR.txt
```diff
@@ -3345,34 +3345,34 @@ STR_6066    :Registra todas as ações do usuário em arquivos no diretório do
...
+STR_6079    :{STRING} demoliu atração ‘{STRING}’.
+STR_6080    :{STRING} mudou aparência da atração ‘{STRING}’.
+STR_6081    :{STRING} mudou o status da atração ‘{STRING}’ para fechado.
+STR_6082    :{STRING} mudou o status da atração ‘{STRING}’ para aberto.
+STR_6083    :{STRING} mudou o status da atração ‘{STRING}’ para testando.
+STR_6084    :{STRING} mudou configurações de veículo da atração ‘{STRING}’.
+STR_6085    :{STRING} mudou configurações de atração da atração ‘{STRING}’.
+STR_6086    :{STRING} renomeou atração de ‘{STRING}’ para ‘{STRING}’.
+STR_6087    :{STRING} mudou preço da atração ‘{STRING}’ para {STRING}
+STR_6088    :{STRING} mudou preço secundário da atração ‘{STRING}’ para {STRING}
```
tupaschoal, line 3366: Seems like the two {STRING} at the end of those lines should be surrounded by simple quotes. They aren't in the original too, so I'm opening for discussion.

tupaschoal (commented):
Thanks for the Rock'n'roll tip, I wouldn't have known.

Also, I noticed I had translated some reference keys on the objects file, we gotta pay attention to that.

data/language/pt-BR.txt
Gymnasiast: I think it is correct. This will be rendered as `Michael mudou preço secundário da atração ‘Woodchip’ para £8,00` - there is no need to put quotes around an amount of currency.

tupaschoal: I see, thanks! Seem like the same applies to the other comment too, so I guess this can be merged as is

objects/pt-BR.json
```diff
@@ -1,19 +1,19 @@
...
   },
   "rct2.nitroent": {
     "reference-name": "'Nitro' Sign",
-    "name": "Placa 'Nitro'"
+    "name": "Placa ‘Nitro’"
   },
   "rct2.walltxgt": {
     "reference-name": "'Texas Giant' Sign",
-    "name": "Placa 'Texas Gigante'"
+    "name": "Placa ‘Texas Gigante’"
```
Gymnasiast, line 16:
```suggestion
    "name": "Placa ‘Texas Giant’"
```

Like “Goliath”, this refers to a sign with the name of the Texas Giant roller coaster.

Issue #2133, opened 2021-02-11
String 6410-6411


telk5093:
New strings from PR OpenRCT2/OpenRCT2#11446.

```
STR_6410    :Zoom in/out
STR_6411    :Show buttons for zooming in and out in the toolbar
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert

Nubbie:
sv-SE already done

PR #2139, opened 2021-02-20
de-DE: Apply #2133, fix some object translations

Wuzzy2:
Applying for issue(s):
- #2133

Also fixes some bad translations in objects file.


de-DE.json
```diff
@@ -0,0 +1,9578 @@
+{
```
Gymnasiast, line 1: Looks like you copied this file by accident.

Wuzzy2:
@Gymnasiast: Whoops! Fixed.

PR #2174, opened 2021-05-07
fi-FI: Update, fix, and complete translations

groenroos:
This PR updates and completes the `fi-FI` translation, by adding in all missing translations (`STR_5553` onwards), ride and scenario descriptions, all 2,151 object translations, as well as various quality and consistency improvements, and translation error fixes.


objects/fi-FI.json
```diff
@@ -1,2537 +1,2537 @@
 {
   "rct2.glthent": {
     "reference-name": "'Goliath' Sign",
-    "name": "'Goliath' Sign"
+    "name": "'Goliath'-kyltti"
```
Gymnasiast, line 4: Please replace the typewriter quotes (`'` and `"`) with whatever it appropriate for Finnish. (Wikipedia page: https://en.wikipedia.org/wiki/Quotation_mark#Finnish_and_Swedish )

groenroos: Good catch - done!

Issue #2181, opened 2021-05-27
Strings 6441-6452


Gymnasiast:
New strings from PRs OpenRCT2/OpenRCT2#10664, OpenRCT2/OpenRCT2#14761 and OpenRCT2/OpenRCT2#14762 .

```
STR_6441    :At least one non-queue footpath surface object must be selected.
STR_6442    :At least one queue footpath surface object must be selected.
STR_6443    :At least one footpath railing object must be selected.
STR_6444    :Footpath Surfaces
STR_6445    :Footpath Railings
STR_6446    :{WINDOW_COLOUR_2}Surface name: {BLACK}{STRINGID}
STR_6447    :{WINDOW_COLOUR_2}Railing name: {BLACK}{STRINGID}
STR_6448    :Unsupported object format
STR_6449    :{WINDOW_COLOUR_2}Tracks:
STR_6450    :{BLACK}“{STRING}”
STR_6451    :{BLACK}“{STRING}” - {STRING}
 STR_6452    :{WINDOW_COLOUR_2}Sells: {BLACK}{STRING}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert

Issue #2165, opened 2021-03-27
String 6436 - 6439


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#14396.

**The `I` in string 6438 stands for “Invisible”. Change as appropriate for your language.**

```
STR_6436    :Toggle invisibility
STR_6437    :Invisible
STR_6438    :I
STR_6439    :Tile Inspector: Toggle invisibility
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert

danidoedel:
@Gymnasiast de-DE is done already ;)

Issue #2158, opened 2021-03-14
String 6435


telk5093:
New strings from PR OpenRCT2/OpenRCT2#14071

```
STR_6435    :{WINDOW_COLOUR_2}Vandals stopped: {BLACK}{COMMA16}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert

danidoedel:
@Gymnasiast de-DE is done already here as well

Wuzzy2:
Deja-vu? I think I already translated this string in a different issue. Strange. It this maybe a mistake?

AaronVanGeffen:
@Wuzzy2 The translation was added by you in #2169. The checkbox just hadn't been ticked, until two hours ago.

Wuzzy2:
I see. It seems I had a brainfart then. :D Thanks.

Issue #2149, opened 2021-03-06
String XXXX


telk5093:
Updated strings on PR OpenRCT2/OpenRCT2#13965.

```
STR_6018    :Construction - Turn left
STR_6019    :Construction - Turn right
STR_6020    :Construction - Use default track
STR_6021    :Construction - Slope down
STR_6022    :Construction - Slope up
STR_6023    :Construction - Toggle chain lift
STR_6024    :Construction - Bank left
STR_6025    :Construction - Bank right
STR_6026    :Construction - Previous track
STR_6027    :Construction - Next track
STR_6028    :Construction - Build current
STR_6029    :Construction - Demolish current
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-OO
- [ ] es-ES
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @Wirlie @JoelTroch @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert

Wuzzy2:
Mark de-DE as completed, please, because it needs no change.

PR #2186, opened 2021-06-15
fr-FR: Apply #2111, #2133, #2140, #2144, #2149, #2158, #2165, #2176, #2181 and improve existing strings

rmnvgr:
Applying for issue(s):
- #2111
- #2133
- #2140
- #2144
- #2149
- #2158
- #2165
- #2176
- #2181

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -3296,18 +3296,18 @@ STR_6014    :Les visiteurs ne paieront que les tickets pour entrer dans les attr
 STR_6015    :Incliné
 STR_6016    :Modifier case
 STR_6017    :Merci de ralentir un peu
-STR_6018    :Construction d’attraction - Tourner à gauche
-STR_6019    :Construction d’attraction - Tourner à droite
-STR_6020    :Construction d’attraction - Utiliser la voie par défaut
-STR_6021    :Construction d’attraction - Pente descendante
-STR_6022    :Construction d’attraction - Pente ascendante
-STR_6023    :Construction d’attraction - Activer/désactiver la remontée à chaîne
-STR_6024    :Construction d’attraction - Incliner à gauche
-STR_6025    :Construction d’attraction - Incliner à droite
-STR_6026    :Construction d’attraction - Morceau précédent
-STR_6027    :Construction d’attraction - Morceau suivante
-STR_6028    :Construction d’attraction - Construire l’actuel
-STR_6029    :Construction d’attraction - Démolir l’actuel
+STR_6018    :Construction - Tourner à gauche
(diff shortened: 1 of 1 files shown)


data/language/fr-FR.txt
```diff
@@ -3643,20 +3645,63 @@ STR_6392    :Impossible de trouver {STRING} à cet emplacement.
...
-STR_6398    :Le fichier contient des types d'attractions non-supportés. Veuillez mettre à jour OpenRCT2.
-STR_6399    :OpenRCT2 a besoin des fichiers originaux de RollerCoaster Tycoon 2 pour fonctionner. Veuillez indiquez le chemin où vous avez installé RollerCoaster Tycoon 2 dans la variable "game_path" dans le fichier config.ini, puis redémarrez OpenRCT2.
-STR_6400    :J'ai téléchargé l'installeur GOG hors-ligne de RollerCoaster Tycoon 2, mais je ne l'ai pas encoré installé
-STR_6401    :J'ai déjà installé RollerCoaster Tycoon 2
+STR_6396    :Désactiver écr. de veille et éco. d’énergie du moniteur
+STR_6397    :Si coché, l’écran de veille et d’autres fonctionnalités d’économie d’énergie du moniteur seront désactivés pendant qu’OpenRCT2 est lancé.
+STR_6398    :Le fichier contient des types d’attractions non supportés. Veuillez mettre à jour OpenRCT2 vers une version plus récente.
+STR_6399    :OpenRCT2 a besoin des fichiers originaux de RollerCoaster Tycoon 2 pour fonctionner. Veuillez indiquer le chemin où vous avez installé RollerCoaster Tycoon 2 dans la variable “game_path” dans le fichier config.ini, puis redémarrez OpenRCT2.
```
Gymnasiast, line 3651: I thought French used guillemets. Or is that only for quotations?

rmnvgr: Guillemets are for quotations, for variables, italics are preferred, but this is impossible here. I used the same quotation marks that are used elsewhere in the translation, but I can replace all of them with guillemets.

Gymnasiast: Well, if guillemets are only for quotations, then I’m not going to ask you to use them here. After all, I want the translation to be correct for French. 

We don’t support italics, indeed. Instead, you could use {SMALLFONT}game_path{MEDIUMFONT}. It will print the text in another font. But I will leave it up you to decide if that is an improvement or if sticking with “” is best after all.

rmnvgr: I may use `{SMALLFONT}` indeed. The original RCT2 translation uses double quotations instead of guillemets, so I'll change them too.

PR #2194, opened 2021-07-20
ko-KR: Improve some strings

telk5093:
1. It might mislead for some Korean that it is unable to demolish the rides/attractions (but can modify them). So I make it clearly.
2. Found un-unified phrases for transparent stations and fixed them

### DIFF
--- a/data/language/ko-KR.txt
+++ b/data/language/ko-KR.txt
@@ -2408,7 +2408,7 @@ STR_3308    :{WINDOW_COLOUR_2}흥미도:
 STR_3309    :{WINDOW_COLOUR_2}{COMMA16}
 STR_3310    :{WINDOW_COLOUR_2}{LENGTH}
 STR_3311    :{WINDOW_COLOUR_2}{COMMA2DP32}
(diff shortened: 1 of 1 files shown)


objects/ko-KR.json
```diff
@@ -5669,11 +5669,11 @@
   },
   "openrct2.station.noentrance": {
     "reference-name": "No entrance",
-    "name": "입구 없음"
+    "name": "출입구 감추기"
```
Gymnasiast, line 5672: The object strings will be redumped soon, which would overwrite your improvements. Could you hold back these changes for now?

telk5093: Sure why not ;)

objects/ko-KR.json
```diff
@@ -5673,7 +5673,7 @@
   },
   "openrct2.station.noplatformnoentrance": {
     "reference-name": "No entrance, no platform",
-    "name": "감추기, 승강장도 감추기"
+    "name": "출입구 및 승강장 감추기"
```
Gymnasiast, line 5676: Hold back this one too :-)

PR #2196, opened 2021-07-27
Notify groenroos for fi-FI

groenroos:
Add myself as a maintainer for `fi-FI`, and for notifications in new/updated string issues.


tupaschoal:
Thanks and welcome @groenroos !

PR #2227, opened 2021-10-01
de-DE: Fix incorrect string ID

Wuzzy2:
I have made a terrible mistake! I accidentally put in the incorrect string ID in my previous PR. This will fix it.

Whoops!


Gymnasiast:
Ah well, it’s only a typo, it’s hardly terrible ;-)

PR #2273, opened 2022-01-21
update sv-SE

antonwe:
added some missing strings that the translation check complained about

Translating key names would only make things more confusing for swedes so I left them unchanged (except alt2/alt gr which is physically different on the keyboard)

### DIFF
--- a/data/language/sv-SE.txt
+++ b/data/language/sv-SE.txt
@@ -3649,8 +3649,8 @@ STR_6446    :{WINDOW_COLOUR_2}Ytnamn: {BLACK}{STRINGID}
 STR_6447    :{WINDOW_COLOUR_2}Räckesnamn: {BLACK}{STRINGID}
 STR_6448    :Okänt objektformat
 STR_6449    :{WINDOW_COLOUR_2}Spår:
-STR_6450    :{BLACK}“{STRING}”
-STR_6451    :{BLACK}“{STRING}” - {STRING}
+STR_6450    :{BLACK}”{STRING}”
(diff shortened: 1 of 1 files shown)


data/language/sv-SE.txt
```diff
@@ -3621,7 +3640,19 @@ STR_6437    :Synlig
...
+STR_6441    :Minst ett icke-kö gångvägsobjekt måste markeras
+STR_6442    :Minst ett kögångvägsobjekt måste markeras
+STR_6443    :Minst ett gångvägsräckesobjekt måste markeras
+STR_6444    :Gångvägsytor
+STR_6445    :Gångvägsräcken
+STR_6446    :{WINDOW_COLOUR_2}Ytnamn: {BLACK}{STRINGID}
+STR_6447    :{WINDOW_COLOUR_2}Räckesnamn: {BLACK}{STRINGID}
+STR_6448    :Okänt objektformat
+STR_6449    :{WINDOW_COLOUR_2}Spår:
+STR_6450    :{BLACK}“{STRING}”
```
Gymnasiast, line 3652: Are you sure this is correct for Swedish? According to https://en.wikipedia.org/wiki/Quotation_mark#Finnish_and_Swedish , Swedish uses a different style of quotation marks.

antonwe: > Are you sure this is correct for Swedish? According to https://en.wikipedia.org/wiki/Quotation_mark#Finnish_and_Swedish , Swedish uses a different style of quotation marks.

Did not realize that those needed to be changed, thanks

PR #2275, opened 2022-01-27
sv-SE: corrected name of twister roller coaster

antonwe:
This was translated to the exact same as the Hypercoaster (STR_6105) for some reason. I've changed it to its original name from the Swedish version of RCT2

### DIFF
--- a/data/language/sv-SE.txt
+++ b/data/language/sv-SE.txt
@@ -55,7 +55,7 @@ STR_0049    :Spökhus
 STR_0050    :Första hjälpen
 STR_0051    :Cirkus
 STR_0052    :Spöktåg
-STR_0053    :Hyperberg- och dalbana
+STR_0053    :Twistberg- och dalbana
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Can you also correct 6109? That one is `Hyper-Twister` in en-GB, so I expect it to be along the lines of `Hyper-Twistberg- och dalbana`. However, it’s currently translated as `Hyperberg- och dalbana med inversioner`, which is completely wrong (as it’s not the hypercoaster but the hyper-_twister_, and because the hyper-twister does not allow inversions.)

antonwe:
> Can you also correct 6109? That one is `Hyper-Twister` in en-GB, so I expect it to be along the lines of `Hyper-Twistberg- och dalbana`. However, it’s currently translated as `Hyperberg- och dalbana med inversioner`, which is completely wrong (as it’s not the hypercoaster but the hyper-_twister_, and because the hyper-twister does not allow inversions.)

I can indeed

Issue #2264, opened 2022-01-17
String 6437+6438


Gymnasiast:
The meaning of these strings have been reversed.

```diff
-STR_6437    :Invisible
-STR_6438    :I
+STR_6437    :Visible
+STR_6438    :V
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert @groenroos @dimateos @rmnvgr

Wuzzy2:
To be honest, it would have better to use 2 new strings instead because I doubt all languages will be updated soon. Reversing the meaning is nasty. But this is too late now, I guess. It's like you're pulling the rug under the translators.

I think it would be best to remove the strings for all languages in which it was not yet translated so the reverse (and VERY wrong) meaning is no longer present in the files.

PR #2276, opened 2022-02-12
Remove outdated translation of STR_6437, STR_6438

Wuzzy2:
This removes the (outdated) translations of `STR_6437` and `STR_6438`. I do this because the meaning of these strings changed to the opposite (invisible to visible), so this is pretty nasty to still have the old (and VERY wrong translations. I therefore remove them in this PR.

See also: #2264


Gymnasiast:
Two of these languages got an update by their translators in the meantime (de and zh-TW) and the remaining three have been updated here: https://github.com/OpenRCT2/Localisation/commit/5310e37c09bdd9ce3032dda63fac25aa9b8b3d00

With that, the old "invisible" translation is now gone from all languages.

PR #2293, opened 2022-03-18
[fr-FR] Improve translation for 'good value'

ldanet:
This is a proposal for an improvement of the translation for `X is really good value`.
The current translation sounds awkward to me (that's personal taste) but more importantly in my view it doesn't convey what this guest's thought is really about: pricing. `X est un très bon choix` sounds to me like the relative quality of the thing is good, but I don't know what is good about it.
`X est vraiment une bonne affaire` is more clearly about the price which would make it a more easily actionable input for the player.
An alternative would be `X est vraiment bon marché`. What would sound most natural to me would be `X est vraiment pas cher` but then you'd have to know what it is to choose between `cher / chère / chers / chères` and there's no way to know what players name their rides.

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -878,7 +878,7 @@ STR_1486    :« Je n’ai pas encore fini mon {STRINGID} »
 STR_1487    :« Le seul fait de regarder {STRINGID} me rend malade »
 STR_1488    :« Je ne paierai pas si cher pour aller sur {STRINGID} »
 STR_1489    :« Je veux rentrer chez moi »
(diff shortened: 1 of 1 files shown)


Gymnasiast:
@JoelTroch Could you review?

You can converse in French.

rmnvgr:
« X est vraiment une bonne affaire » sounds good to me :+1:

JoelTroch:
I agree with @rmnvgr & @ldanet .

@Gymnasiast (or any other OpenRCT2 developer/maintainer): this PR can be merged.

PR #2325, opened 2022-05-02
Update 6438 for all languages

Broxzier:
The column header of the tile inspector's visibility column has been updated to an eye glyph.


Gymnasiast:
Oh, this is something I missed. I’d much prefer keeping the original letter, actually. It seems much more consistent with the rest. For example, the "last on tile" flag has a tick, but the column has the letter `L` on it.

Broxzier:
This change was specifically asked for on the PR. Personally, I think the column title could be removed entirely (and the tool-tip could maybe even be moved to the entire column instead of just the header). The eyes in front of the rows make its function intuitive, especially considering they are there by default. Other tools that use eye symbols for visibility don't have a header for that column (Unity, Blender, Photoshop).

We could ask the community what they prefer.

Gymnasiast:
Seems the community had a slight preference for this. I wasn’t going to die on the hill, so I will merge it.

PR #2326, opened 2022-05-09
zh-CN: Apply Updates

LPR-Michael:
…n zh-CN.json

With some new update, Some new content are not translate, So I fixed it.


tupaschoal:
Why did you close @LPR-Michael ? If it's work in progress, you can keep it as draft and notify us when it's available :)

LPR-Michael:
I translated all string in zh_CN.txt and zh_CN.json to match new update.
P.S. I'm not familiar with GitHub, So thank you for helping my to set right state for this pull requests.

PR #2338, opened 2022-05-15
ca-ES: Provide Catalan localisation for several objects.

J0anJosep:
This PR provides ~50 % of the object strings translated into Catalan.

### DIFF
--- a/objects/ca-ES.json
+++ b/objects/ca-ES.json
@@ -1,621 +1,621 @@
 {
   "rct2.glthent": {
     "reference-name": "'Goliath' Sign",
-    "name": "'Goliath' Sign"
+    "name": "Senyal «Goliath»"
   },
   "rct2.mdsaent": {
     "reference-name": "'Medusa' Sign",
-    "name": "'Medusa' Sign"
+    "name": "Senyal «Medusa»"
   },
   "rct2.nitroent": {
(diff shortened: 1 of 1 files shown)


objects/ca-ES.json
```diff
@@ -1,1223 +1,1223 @@
...
   },
   "rct2.tt.gangster": {
     "reference-name": "Animatronic Gangster",
-    "name": "Animatronic Gangster"
+    "name": "Animatrònic de gànster"
   },
   "rct2.tt.gdalien2": {
     "reference-name": "Animatronic Good Alien",
-    "name": "Animatronic Good Alien"
+    "name": "Animatrònic d'alienígena bo"
```
Gymnasiast, line 360: Please use a proper apostrophe instead of the typewriter one. This also applies to other occurrences.

```suggestion
    "name": "Animatrònic d’alienígena bo"
```

J0anJosep: Thank you. I will look into it.

By the way, is it possible that some quotation marks in en-GB are wrong?

imatge

Are they ok? Should I report a bug?

Gymnasiast: Those are indeed not OK. You don’t need to report it, I will fix it in the repository straight away.

PR #2352, opened 2022-05-28
es-ES: Consinstency + typos

dimateos:
Checked a bit of the strings and fixed some issues:
* **Consistent dot endings** (although in English and other languages are not present). In Spanish they follow the same rules, but for consistency with other strings I guess for now is best to just stick with them.
* **Non es-ES words** (I think they were argentinean): Monorriel, Manita, Calesita...
* Added **ordering sense** to the strings STR_0779 like in English and other languages (they were just numbers)
* Some cut-short strings that were incomplete. _There might be some more_.

I saw some missing strings (present in other languages like fr-FR, ru-RU)
* For example STR_0532 and some around it.
* I might check them and create a issue / add directly.
* _They might be missing in other languages too?_.


data/language/es-ES.txt
```diff
@@ -150,73 +150,73 @@ STR_0564    :Esta montaña rusa, construida en madera, es rápida, rústica, rui
...
-STR_0801    :23
-STR_0802    :24
-STR_0803    :25
-STR_0804    :26
-STR_0805    :27
-STR_0806    :28
-STR_0807    :29
-STR_0808    :30
-STR_0809    :31
+STR_0779    :1º
```
Gymnasiast, line 189: These are used for dates. Would you use `1º` in dates, or just `1`?

dimateos: Sorry for the delay. Yeah you might be right. 1º is possible but nowdays we just use 1.

PR #2383, opened 2022-07-24
ja-JP: Apply #2176 + miscellaneous

johnwdolph:
Applying for issue(s):
- #2176


Gymnasiast:
@AaronVanGeffen

data/language/ja-JP.txt
```diff
@@ -32,12 +32,12 @@ STR_0027    :バンパーカー
 STR_0028    :パイレーツシップ
 STR_0029    :スイング・インバータ・シップ
 STR_0030    :食べ物の店
-STR_0031    :Unknown Stall (1D)
+STR_0031    :未知の売店 (1D)
```
AaronVanGeffen, line 35: Any particular reason you went with 未知 for this?

johnwdolph: > Any particular reason you went with 未知 for this?

Good point. There is no particular reason.

Would something like this be more accurate? 

```suggestion
STR_0031    :不詳な売店 (1D)
```
Those translations can be discarded or modified.

johnwdolph: @AaronVanGeffen

> Any particular reason you went with 未知 for this?

That is how I understood the phrase to work.
I removed those changes since I am not 100% certain it is correct.

---

I did some research. This might be better:

不詳な売店

Instead of: 

未知の売店

Gymnasiast:
@johnwdolph Looks like you tried to update the branch, but went wrong somewhere. Could you rebase/fix it?

data/language/ja-JP.txt
```diff
@@ -1856,7 +1856,7 @@ STR_2521    :スタッフ名鑑の表示
 STR_2522    :最近のメッセージを見る
 STR_2523    :地図を見る
 STR_2524    :スクリーンショット
-STR_2533    :Backspace
+STR_2533    :バックスペース
```
AaronVanGeffen, line 1859: I don't think this should be transliterated into katakana. On a [Japanese keyboard](https://en.wikipedia.org/wiki/Keyboard_layout#/media/File:KB_Japanese.svg) the key still reads 'backspace'.

AaronVanGeffen (approved):
Disagree with the transliteration for string 2533. The other three strings are fine with me.

data/language/ja-JP.txt
Gymnasiast:
```suggestion
STR_2533    :Backspace
```

PR #2409, opened 2022-08-05
Apply 17328's map size change to all languages

Broxzier:
This applies a static language line change from https://github.com/OpenRCT2/OpenRCT2/pull/17328 to all language files.


Gymnasiast:
It passed me by on the actual PR, but in the future, it’s best to just create a new string and decommission the old one if the contents are incompatible or otherwise significantly changed. This ensures that other strings remain working and is also a good thing if/when we start using more refined tools to maintain translations.

Broxzier:
I wouldn't say this is a significant change, this string entry's entire use is literally for the park size fields. Ideally, fields like this shouldn't be part of the language string entries, but rather something that tells the game how to render numbers for the specified language (something we now do with UINT16/COMMA16).

Gymnasiast:
It is a significant change as it breaks the display until that language is updated.

PR #2412, opened 2022-08-09
ca-ES: Change verbal periphrases.

J0anJosep:


data/language/ca-ES.txt
```diff
@@ -493,7 +493,7 @@ STR_1102    :Movent-se a {VELOCITY}
 STR_1103    :Arribant a {POP16}{STRINGID}
 STR_1104    :Descarregant passatgers en {POP16}{STRINGID}
 STR_1105    :Viatjant a {VELOCITY}
-STR_1106    :Està xocant!
+STR_1106    :Ei! Que xoca!
```
Gymnasiast, line 496: This seems to deviate quite a lot from the actual meaning. In English, this says “Crashing!”, and it is displayed as a status when a roller coaster train has derailed and is in the process of crashing (like an aeroplane). Similarly, String 1107 is the status when this crash has actually happened. You can see it for yourself by building a powered launched coaster and making the launch speed too high.

J0anJosep: @Gymnasiast Thanks for reviewing it so thoroughfully! I know what those strings are used for, but I find it difficult to translate them. I think the previous suggestion was OK anyway.
In case you tried Google Translate, it won't translate "que xoca" the right way ("what a shock"; that translation is absolutely wrong). If you give a little bit of context, like in "un vehicle que xoca", it will: "a crashing vehicle".
Thank you for letting me know.

Gymnasiast: No problem, I try to make the translations as good as they can be. I did indeed use Google Translate, which is why I was a bit careful and said “it _seems” ;-)

PR #2440, opened 2022-09-02
de-DE: Add TTPIRF translations

Wuzzy2:
This adds the missing German translations for the TTPIRF strings


Gymnasiast:
Sorry for this, but translations for JSON objects are in the https://github.com/OpenRCT2/Localisation/blob/master/objects/de-DE.json file. The section where you added the translations is for overriding .DAT objects only. 

This section will be removed eventually, though. Also, I’m working on a proper translator, as the current workflow is not just awkward for the translators, but also for us.

Issue #2446, opened 2022-09-25
STR_6519


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#18084.

```
STR_6519    :Extras
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @daihakken

Issue #2430, opened 2022-08-26
String 96 and 606


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#17900.

```
 STR_0096    :Classic Wooden Roller Coaster
STR_0606    :An older-style wooden roller coaster with a fast and rough ride, with plenty of air-time, some lateral G’s, and designed to feel ‘out-of-control’
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @daihakken

PR #2448, opened 2022-09-26
pt-BR: Apply #2430, #2446

tupaschoal:
Applying for issue(s):
- #2430
- #2446


data/language/pt-BR.txt
```diff
@@ -4065,9 +4072,24 @@ STR_DTLS    :
 ###########
 
 ## Start OpenRCT2 Official
+[TTPIRF02]
```
Gymnasiast, line 4075: These were removed deliberately, as they have been made into JSON objects.

tupaschoal: I thought there was something weird, but the check was pointing out they were missing and I checked en -GB and they seem to be there

tupaschoal: Done.

PR #2455, opened 2022-10-01
ca-ES: Amend some strings

J0anJosep:
- Better use of preposition "per": changed in some places with "per a", "per tal que" and other variations.
- Change "escollir" with "triar" in some cases.
- Add a space between number and % symbol (Catalan rules about this are not the same as in English; I am not using a non-breaking space as I don't know whether OpenRCT2 will handle it right).
- Replace more uses of gerunds instead of the more appropriate present tense.
- Other small amends.


Gymnasiast:
OpenRCT2 will handle a non-breaking space right, so please use it.

Issue #2369, opened 2022-07-04
String XXXX


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#17499.

```
STR_6499    :Vehicle type not supported by track design format
STR_6500    :Track elements not supported by track design format
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert @groenroos @dimateos @rmnvgr

PR #2465, opened 2022-10-05
String 6499,6500

fernandosutter:
Applying for issue(s):
- #2369


Gymnasiast:
Could you use the form `ar-EG: Apply #2369` for any future PRs and commits? It helps us keep track of translation status more easily.

PR #2489, opened 2022-10-27
zh-CN: Apply #2473

LPR-Michael:
Applying for issue(s):
- #2473

And some change for better display


Gymnasiast:
This does not seem to apply #2473, it only adds some newlines to existing strings.

Issue #2509, opened 2022-11-23
Strings 6530-6532 + Scenario descriptions


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#18607.

```
STR_6530    :User Created Expansion Set
STR_6531    :The Time Machine (2003)
STR_6532    :Katy’s Dreamworld (2003)

###
<UCES Halloween>
STR_SCNR    :Cemetery Ridge
STR_PARK    :Cemetery Ridge
STR_DTLS    :This is Halloween, UCES Halloween, pumpkins scream in the dead of night! This graveyard is in trouble and it’s up to you to save it, while letting the dead rest in peace! Can you keep the ghosts in their graves and bring chills to your customers?{NEWLINE}Author: Squid

###############################################################################
## UCES: The Time Machine (2003)
###############################################################################
<Lighthouse of Alexandria by Katatude for UCES>
STR_SCNR    :The Lighthouse of Alexandria
STR_PARK    :The Lighthouse of Alexandria
STR_DTLS    :Alexander built the city; Greeks, Romans, Egyptians left their mark. But the biggest honor was a Wonder of the Ancient World - the Lighthouse. Visit and make a park!{NEWLINE}Author: Katatude

<Cleveland's Luna Park>
STR_SCNR    :Luna Park, Cleveland
STR_PARK    :Luna Park
STR_DTLS    :As it was on its opening day - 18 May 1905.{NEWLINE}Author: Aetherwave

<Mount Vesuvius 1700 A.D. by Katatude for UCES>
STR_SCNR    :Mount Vesuvius
STR_PARK    :Mount Vesuvius
STR_DTLS    :Pompeii and Herculaneum were buried by Mt. Vesuvius in 79 A.D. Visit the excavations and build a park!{NEWLINE}Author: Katatude

<The Sandbox by Katatude for UCES>
STR_SCNR    :The Sandbox
STR_PARK    :The Sandbox
STR_DTLS    :What everyone wants - a sandbox! So grab your l’il pail and shovel and build a park!{NEWLINE}Author: Katatude

<Niagara Falls & Gorge by Katatude for UCES>
STR_SCNR    :Niagara Falls & Gorge
STR_PARK    :Niagara Falls
STR_DTLS    :American Falls, Bridal Falls & Canadian Horseshoe Falls on the Niagara Frontier, 1850.{NEWLINE}Author: Katatude

<Rocky Mountain Miners>
[...]

PR #2510, opened 2022-11-23
ko-KR: Apply #2509

telk5093:
Applying for issue(s):
- #2509 

I decided to translate ``STR_6530`` as I commented at [#2509.](https://github.com/OpenRCT2/Localisation/issues/2509#issuecomment-1324925260)
(btw, UCES is not very known in South Korea)


Gymnasiast:
> I decided to translate STR_6530 as I commented at https://github.com/OpenRCT2/Localisation/issues/2509#issuecomment-1324925260

That is fine. You have the freedom to do things like that if it makes for a better Korean localisation.

Issue #2511, opened 2022-11-23
String 5836, 6515, 6516, 6517


Gymnasiast:
Updated strings on https://github.com/OpenRCT2/OpenRCT2/commit/8e85c2f344b6cd894ab898f823972f80e76388d0.

STR_5836 referenced the incorrect tab. The other strings only had "RCT1" changed to "RollerCoaster Tycoon 1" for clarity.

```diff
-STR_5836    :Select music to use on the main menu.{NEWLINE}Selecting the RCT1 theme requires that you set the path to RCT1 in the Miscellaneous tab.
-STR_6515    :{BLACK}RCT1 not linked - fallback images will be used.
-STR_6516    :One or more objects added require RCT1 linked for proper display. Fallback images will be used.
-STR_6517    :One or more objects in this park require RCT1 linked for proper display. Fallback images will be used.
+STR_5836    :Select music to use on the main menu.{NEWLINE}Selecting the RCT1 theme requires that you set the path to RCT1 in the Advanced tab.
+STR_6515    :{BLACK}RollerCoaster Tycoon 1 not linked - fallback images will be used.
+STR_6516    :One or more objects added require RollerCoaster Tycoon 1 linked for proper display. Fallback images will be used.
+STR_6517    :One or more objects in this park require RollerCoaster Tycoon 1 linked for proper display. Fallback images will be used.
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [ ] cs-CZ
- [x] da-DK
- [x] de-DE
- [ ] en-US
- [x] eo-ZZ
- [x] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] hu-HU
- [x] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [x] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [x] zh-CN
- [x] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @daihakken

Issue #2492, opened 2022-11-08
String XXXX


Gymnasiast:
Updated strings on [PR OpenRCT2/OpenRCT2#0000](https://github.com/OpenRCT2/OpenRCT2/commit/bfe7168b5611409dbd35e0dd891c70775338fdf3).

```diff
-STR_6322    :{WINDOW_COLOUR_2}Sprite Id: {BLACK}{INT32}
+STR_6322    :{WINDOW_COLOUR_2}Entity ID: {BLACK}{INT32}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @daihakken

Issue #2472, opened 2022-10-11
Strings 6520-6526


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#18050.

```
STR_6520    :Asset Packs
STR_6521    :Low Priority
STR_6522    :High Priority
STR_6523    :Decrease the priority of the selected asset pack.
STR_6524    :Increase the priority of the selected asset pack.
STR_6525    :Reload all assets in the game with the enabled asset packs.
STR_6526    :(base graphics, music and sound effects)
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @daihakken

PR #2513, opened 2022-11-25
de-DE: Apply #2472, #2473, #2492, #2509, #2511

Wuzzy2:
Applying for issue(s):
* #2472
* #2473
* #2492
* #2509
* #2511


tupaschoal:
@Wuzzy2 I see `STR_6530` was left out of this review. I know it's not supposed to be translated, but I think it's better if you replicate it on `de-DE` for better results on the lint checking.

PR #2543, opened 2022-12-25
Fixing some translation errors in Polish localization

Patrxgt:
Fixing some translation errors in Polish localization, like names of the options too long to fit in the window they supposed to fit in or incorrect spelling in autosave interval option.


Gymnasiast:
I cannot merge it like this I’m afraid - you first deleted the file and then uploaded an improperly formatted one (probably using DOS line endings instead of Unix ones), which means all 4371 lines are now changed. You’ll have to fix that.

PR #2548, opened 2023-01-03
Create fr-CA to the languages

TheClaudeQC:
like this? Because I can't really put the .txt object

Applying for issue(s):


TheClaudeQC:
is everything fine, dose a code is needed to rollback to original French?

TheClaudeQC (commented):
is a rollback to fr-FR needed?

Gymnasiast:
I merged the file here: https://github.com/OpenRCT2/Localisation/commit/cdbcc49e79da34d611766f162dc9b8806fd3c9e2

I will take care of adding it to the menu and setting up the fallback to European French. You only have to update the file with strings _that are different to European French_.

Issue #2558, opened 2023-01-19
Tycoon Park


Gymnasiast:
**Add these translations under those for "Panda World". Do not translate the part between the < > brackets.**

```
<Tycoon Park>
STR_SCNR    :Tycoon Park
STR_PARK    :Tycoon Park
STR_DTLS    :
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @LPR-Michael @daihakken

tygrysek90:
cs-CZ no translation needed

Given names of places, parks, companies etc. are generally accepted in the original form (as in language-wide and as in OpenRCT2). 

JSON data structure found well at [cs-CZ.json @ line 3416](https://github.com/OpenRCT2/Localisation/blob/master/objects/cs-CZ.json#L3416) .

_Note: One of few given names that have a Czech translation is "Netherlands", which is called "Nizozemí"_

PR #2560, opened 2023-01-21
zh-CN: Apply #2558 #2562 #2563 + Some fixes for better display and some errors

LPR-Michael:
Applying for issue(s):
- #2558
- #2562
- #2563

Some fixes for better display and some errors:
Unified translation for same word.
Move UCES part of Scenarios to the place same as en-GB.txt.
Replace some STR which use too many Internet buzzwords that already out of date.
Replace all punctuation.


data/language/zh-CN.txt
```diff
@@ -246,15 +246,15 @@ STR_0833    :暂停游戏
 STR_0834    :磁盘和游戏选项
 STR_0839    :{UINT16} × {UINT16}
 STR_0840    :{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{UINT16} × {UINT16}
-STR_0847    :关于‘OpenRCT2’
+STR_0847    :关于'OpenRCT2'
```
Gymnasiast, line 249: Please don’t use typewriter/straight quotes. Use whatever is appropriate for Simplified Chinese. (See e.g. https://en.wikipedia.org/wiki/Quotation_mark )

LPR-Michael: Now should be OK.

LPR-Michael:
Now should be OK. @Gymnasiast

PR #2596, opened 2023-03-30
Fix some German typos

Wuzzy2:
This PR fixes some typos and mistakes of the German translation. Also includes string parameters that were overlooked for some reason.


tupaschoal (approved):
Great find about the missing parameters!

PR #2617, opened 2023-04-11
translation_check: Report repeated and unexpected keys

tupaschoal:
We had a Python and a PowerShell script doing roughly the same thing, but the PowerShell one was not being watched nor maintained, so I removed it. 

There were some checks there that we were not doing, so I took the chance of incorporating those into the Python script, but all of the whitespace stuff I just left out


tupaschoal:
@ZehMatt I'm deleting its yml here, not sure if that will be enough or if @IntelOrca will have to go to appveyor website. In any case, purging it is definitely my intention.

ZehMatt (approved):

Gymnasiast (approved):

Issue #2609, opened 2023-04-06
Strings 6551-6576


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#19446.

```
STR_6551    :Army green
STR_6552    :Honeydew
STR_6553    :Tan
STR_6554    :Maroon
STR_6555    :Coral pink
STR_6556    :Forest green
STR_6557    :Chartreuse
STR_6558    :Hunter green
STR_6559    :Celadon
STR_6560    :Lime green
STR_6561    :Sepia
STR_6562    :Peach
STR_6563    :Periwinkle
STR_6564    :Viridian
STR_6565    :Seafoam green
STR_6566    :Violet
STR_6567    :Lavender
STR_6568    :Pastel orange
STR_6569    :Deep water
STR_6570    :Pastel pink
STR_6571    :Umber
STR_6572    :Beige
STR_6573    :Invisible
STR_6574    :Void
STR_6575    :Allow special colour schemes
STR_6576    :Adds special colours to colour dropdown
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @LPR-Michael @daihakken

PR #2639, opened 2023-04-21
de-DE: Apply #2609

loeppenberg:
Applying for issue(s):
- #2609


loeppenberg:
The three 'Same' translations are:
'Chartreuse', 'Sepia' and 'Beige'
'Chartreuse' - I personally have never heard of that colour, but this french word [seems to be used in german just as it is](https://farbtonkarte.de/farbe-chartreuse-hex-rgb-cmyk/) used in english for this yellow-greenish colour.
'Sepia' and 'Beige' - both are common in the german language as is.

PR #2700, opened 2024-01-06
de-DE: fix a typo and improve formatting consistency

Sadret:
- changed "rauh" to "rau", which is the correct spelling since 1996
- removed the full stop at the end of some sentences, where there is no full stop in similar sentences


Gymnasiast:
@Sadret In the Dutch translation, I used a full stop at the end of every description as I considered that more in line with how Dutch is written.

 I don’t know if that also applies to German - I’ll leave that to you.

Sadret:
Right now, there are three ride description with and all the others without. I just wanted to make that consistent, without questioning which one makes more sense.

In my opinion, we should have full stops there, too. But I also do not see it as very important. Also, some of the descriptions are not sentences in the grammatical sense:
"A compact roller coaster with individual cars and smooth twisting drops"
"Eine kompakte Achterbahn mit unabhängigen Wagen und gemäßigten, sich windenden Gefällen"

Gymnasiast:
@Sadret Feel free to modify these. The German translation has no maintainer at the moment and improvements are always welcome.

PR #2705, opened 2024-01-06
de-DE: adjust translation of STR_6515

Sadret:
The sentence should not contain in upper case after the hyphen. Either it must continue in lower case or the hyphen can be changed to a full stop.
- Argument for the former is that it is closer to the english sentence.
- Argument for the latter is that it is more consistent with STR_6516 and STR_6517.
- A (subjective) choice was made for the latter.


Gymnasiast:
> Argument for the former is that it is closer to the english sentence.

How the original English expresses something is not relevant. The only thing that matters is what is being communicated. Always pick the thing that makes the most sense in the target language.

PR #2716, opened 2024-01-18
de-DE: remove abbreviation in 1673

Sadret:
It does not make sense to abbreviate only two letters. Space-wise, the english version is much longer.


Gymnasiast:
It may have been necessary in the old situation, where that box was narrower.

Issue #2732, opened 2024-02-11
STR_6613


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#21376.

```
STR_6613    :Reload object
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Goddesen @Gymnasiast @marcinkunert @Tupaschoal @LPR-Michael @daihakken

Mmmmulder:
Can I be added to receive notifications when new strings arrive?

Gymnasiast:
> Can I be added to receive notifications when new strings arrive?

Done!

PR #2736, opened 2024-02-12
Update nb-NO.txt

Mmmmulder:
Shortened text that didn't fit windows.
Fixed spelling and other errors.
Added RCT2 scenarios.
Removed music style strings.


Gymnasiast:
Hi, could you also list any issues your PR address? In this case, it applies #2732.

Mmmmulder:
> Hi, could you also list any issues your PR address? In this case, it applies #2732.

right, forgot, that's the one #2732

PR #2741, opened 2024-02-13
Remove old music style strings

tupaschoal:
Applying OpenRCT2/OpenRCT2#21385 to all languages


Gymnasiast:
Ideally, we’d also update the object files.

tupaschoal:
You mean check whether the translations there updated too?

Gymnasiast:
Yeah, I think that some translations might not have been synced.

tupaschoal:
> Yeah, I think that some translations might not have been synced.

I was thinking of doing it myself, but we don't know which is the most accurate, the one on objects or the one on the translation file. I'm thinking about opening an issue on the Localisation repo and have the translators notified to check if they need update, what do you think @Gymnasiast ?

Issue #2345, opened 2022-05-28
Strings 0095, 0605


telk5093:
New strings from PR OpenRCT2/OpenRCT2#16825.

```
STR_0095    :Alpine Coaster
STR_0605    :Riders toboggan down a meandering steel track, braking to control their speed
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @octaroot @LPSGizmo @danidoedel @Wuzzy2 @tellovishous @rzhao271 @anon569 @LucaRed @AaronVanGeffen @telk5093 @Goddesen @Gymnasiast @renansimoes @Tupaschoal @izhangfei @daihakken @marcinkunert @groenroos @dimateos @rmnvgr

PR #2743, opened 2024-02-14
Update nb-NO.txt

Mmmmulder:
Small fix


Gymnasiast:
Could you start putting more descriptive commit titles?

E.g.:
`nb-NO: Small fixes`
or
`nb-NO: Apply #1234, #2345`

rather than just leaving the default of `Update nb-NO.txt`.

I rewrote them up until now as I wanted to get you up to speed gently, but I think you’re ready for this next step now ;)

Mmmmulder:
right, didn't catch that, sorry, thanks for the heads up

PR #2763, opened 2024-02-23
uk-UA: Add as maintained language and track on issues

tupaschoal:
@CsyeCokTheSolly I'm adding `uk-UA` as an active language. Let us know if you want us to add you to the issue notification whenever a new string is added, or if you'd rather wait until the translation is finished.


CsyeCokTheSolly:
Alright, thanks for telling me. Me and my colleagues still intend on translating, now with more people on board.

PR #2777, opened 2024-03-08
en-us Bobsled consistency

ZeeMaji:
The Bobsleigh Coaster ride type is known as the "Bobsled Coaster" in en-us, but this change was not reflected on the vehicles for the ride type. This fixes that.


Gymnasiast:
For any following PRs, could you write the commit message in the form of `en-US: Consistently use “Bobsled” instead of “Bobsleigh”`?

PR #2780, opened 2024-03-11
en-US: Update WWTT flying saucer vehicle names

ZeeMaji:
The en-GB strings for these objects were recently changed on the objects repo to refer to them as "Dodgems" so as such, this updates them to "Bumper Cars" in en-US.


Gymnasiast:
Could you also create an issue here on the Localisation repo for the other languages?

Issue #2815, opened 2024-05-16
Strings 6629–6630


AaronVanGeffen:
New strings from PR OpenRCT2/OpenRCT2#21989.

```
STR_6629    :Align toolbar buttons horizontally centred
STR_6630    :This setting will align toolbar the buttons horizontally in the centre of the screen. The traditional way of aligning them is to in left and right corner.
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken

PR #2816, opened 2024-05-16
zh-CN: Apply #2814, #2815, #2821 & fix ellipsis convention

mrmagic2020:
Applying for issue(s):
- #2814 
- #2815 
- #2821


AaronVanGeffen:
You seem to be doubling the amount of ellipses. Is this intentional?

mrmagic2020:
> You seem to be doubling the amount of ellipses. Is this intentional?

Yes. In Chinese, the standard way of writing ellipses is having six dots, instead of three. Further reference: https://en.wikipedia.org/wiki/Ellipsis#:~:text=In%20Chinese%2C%20the%20ellipsis%20is,vertical%20space%20as%20two%20characters).

Gymnasiast:
> > You seem to be doubling the amount of ellipses. Is this intentional?
> 
> Yes. In Chinese, the standard way of writing ellipses is having six dots, instead of three. Further reference: https://en.wikipedia.org/wiki/Ellipsis#:~:text=In%20Chinese%2C%20the%20ellipsis%20is,vertical%20space%20as%20two%20characters).

That page does say the dots are commonly vertically centred - i.e. `⋯⋯`, rather than `……` - is that correct?

mrmagic2020:
> > > You seem to be doubling the amount of ellipses. Is this intentional?
> > 
> > 
> > Yes. In Chinese, the standard way of writing ellipses is having six dots, instead of three. Further reference: https://en.wikipedia.org/wiki/Ellipsis#:~:text=In%20Chinese%2C%20the%20ellipsis%20is,vertical%20space%20as%20two%20characters).
> 
> That page does say the dots are commonly vertically centred - i.e. `⋯⋯`, rather than `……` - is that correct?

Hmmm, interesting. Nevertheless, from my experience, people tend to write ellipses like this `……`. It's also what I get from SHIFT+6 on a Simplified Chinese keyboard... I think both will do as long as they're used consistently 😅

Gymnasiast:
Let’s go with that, then - you know more about Chinese than we do, obviously ;)

Issue #2826, opened 2024-05-19
Strings 6638-6639


mrmagic2020:
New strings from PR OpenRCT2/OpenRCT2#21853.

```
STR_6638    :Enlarged UI
STR_6639    :Modifies the interface to be more suitable for touch usage
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken

Issue #2821, opened 2024-05-17
Strings 6631–6637


AaronVanGeffen:
New strings from PR OpenRCT2/OpenRCT2#21893.

```
STR_6631    :Loading...
STR_6632    :Checking object files...
STR_6633    :Checking scenario files...
STR_6634    :Checking track design files...
STR_6635    :Checking asset packs...
STR_6636    :Checking title sequences...
STR_6637    :Loading title sequence...
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken

Issue #2806, opened 2024-04-30
String 6627


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#21850.

```
STR_6627    :Track speed too high!
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [X] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @daihakken

Issue #2801, opened 2024-04-30
Strings 0097-0607


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#21642.

```
STR_0097    :Classic Stand-up Roller Coaster
STR_0607    :An intense, older-style steel looping roller coaster where the riders ride in a standing position
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [X] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@Omaranwa @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @daihakken

Gymnasiast:
Thanks for reporting, forgot about it :+1:

PR #2834, opened 2024-05-20
Apply #2794, #2801, #2806, #2814, #2815, #2821, #2826

rmnvgr:
Applying for issues:
- #2794
- #2801 
- #2806 
- #2814 
- #2815 
- #2821 
- #2826 

Also fix an old string being too long and going out of the settings window.

### DIFF
--- a/data/language/fr-FR.txt
+++ b/data/language/fr-FR.txt
@@ -100,6 +100,7 @@ STR_0093    :Montagnes russes hybrides
 STR_0094    :Montagnes russes à rail unique
 STR_0095    :Montagnes russes alpines
 STR_0096    :Montagnes russes en bois classiques
+STR_0097    :Montagnes russes debout classiques
 STR_0512    :Des voitures individuelles descendent des pentes régulières en virage
 STR_0513    :Grand huit que les passagers parcourent debout
 STR_0514    :Des voitures suspendues sous la voie se balancent dans les virages des montagnes russes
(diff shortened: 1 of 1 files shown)


data/language/fr-FR.txt
```diff
@@ -3700,6 +3702,23 @@ STR_6619    :Ce type d’objet ne peut pas être restreint !
 STR_6620    :Objet introuvable !
 STR_6621    :Restreindre
 STR_6622    :Restreint l’objet à l’éditeur de scénario et au mode bac à sable
+STR_6623    :Tapez {SMALLFONT}help{MEDIUMFONT} pour une liste des commandes disponibles. Tapez {SMALLFONT}hide{MEDIUMFONT} pour cacher la console.
```
Gymnasiast, line 3705: This will not show up correctly if the console it set to use the small font instead of medium - it will start off as a small font string and switch to medium after the first few words.

I guess the best option is to force the entire string to appear in medium font. Will need to be tested, though.

```suggestion
STR_6623    :{MEDIUMFONT}Tapez {SMALLFONT}help{MEDIUMFONT} pour une liste des commandes disponibles. Tapez {SMALLFONT}hide{MEDIUMFONT} pour cacher la console.
```

rmnvgr: Oh I didn't know the console could use the small font. It'd look strange to only have the help text in medium font, I think it's better to use regular guillemets then.

Gymnasiast (approved):

PR #2842, opened 2024-05-26
Apply OpenRCT2#22065 to all language files

AaronVanGeffen:
This adjusts `STR_5759` and `STR_6378` for all languages, accounting for #22065.

I noticed French and Ukrainian used a localised version of the KiB unit, so I have added a replacement `STR_6643` for both.


Gymnasiast (dismissed):
Please add STR_6642 and STR_6643 to _all_ languages. We never leave out interface translations, even if they are the same. (Only exceptions being en-US and fr-CA, which explicitly extend another language.)

PR #2861, opened 2024-06-23
tr-TR: Fixed typos and added translations for the untranslated lines

ozankaraali:
Dear maintainers,

First of all thanks for maintaining this game. While I was playing, I have seen that Turkish translations had a lot of typos (and probably still there will be after this PR). But still I wanted to improve the translations and I have changed a couple of strings.

I tested in my game and some strings were English, even though I translated to Turkish, I don't know if OpenRCT2 supports unicode completely (I assume it is) but still if an experienced maintainer can take a look why, it would be a huge help.

Long story short, I am proposing an improvement for tr-TR since it is unmaintained for a while (yet still I cannot be a maintainer since I do not have enough time). Feel free to give feedback. Hope this helps.

Best,
Ozan

### DIFF
--- a/data/language/tr-TR.txt
+++ b/data/language/tr-TR.txt
@@ -552,8 +552,7 @@ STR_1167    :Su seviyesi burada yükseltilemez…
 STR_1168    :Seçenekler
 STR_1169    :(Hiç birisi)
 STR_1170    :{STRING}
-STR_1171    :{RED}Kapalı - -
-STR_1172    :{YELLOW}{STRINGID} - -
+STR_1171    :{RED}Kapalı
 STR_1173    :Yollar ve bekleme sıraları
 STR_1174    :Banner engelliyor
 STR_1175    :Eğimli yola koyulamaz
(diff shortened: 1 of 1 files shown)


Gymnasiast:
> I tested in my game and some strings were English, even though I translated to Turkish, I don't know if OpenRCT2 supports unicode completely (I assume it is) but still if an experienced maintainer can take a look why, it would be a huge help.

OpenRCT2 supports Unicode, but the sprite font needs images to be drawn up for every character. For Turkish, I added the missing ones a few years ago.

The reason some strings appear in English is not related to that, though. I looked at the file and currently, it’s translated up until STR_6527. However, the English translations currently goes to 6645, meaning there are almost 120 untranslated strings at the end. Besides that, some objects may lack translations. These are saved in another file in this repository.

ozankaraali:
Thanks @Gymnasiast, maybe my game did not load correctly but it was lacking even the translation of "Haunted House" which the prior translation file renders. But you are correct, I found in the objects/tr-TR.json, Haunted House is in English. Either way, I believe this one fixes the most of the typos so it should be a better translation overall, on top of the existing. If I will have time later, I can try to extend that and thanks for merging and heads up about the objects.

Have a nice day,
Ozan

Issue #2866, opened 2024-06-28
STR_2386 doesn't show correct month/year in some languages. 


yanleiberg:
For all scenario goal with a guest number requirement and park rating 600 in rct1 and rct2, 
it always displays the deadline to " the end of March, Year 1, with a park rating of at least 600“ in Chinese and Japanese

string id: STR_2386    

1. wrong {monthyear} displaying for the scenario goal string (STR_2386) in zh_cn, zh_tw, ja-JP
2. unnecessary {newline} before number of guest in zh_cn

surprisingly this is correctly displaying in ko-KR.
STR_2387 is correct now in zh_cn, zh_tw, ja-JP, please **don't** modify it. 

How to fix: 
I believe we should simply remove the “{POP16}” before {MONTHYEAR} in STR_2386 but I am not sure for that. 

I am still new to OpenRCT2 even I love RCT1/2, I would like to contribute if possible. 

Here are 2 examples: 

---Forest Frontiers---

  actual display in game: 
	在第1年3月末乐园内至少有
	250游客，并且乐园评分不低于600

  translated from actual display: 
	To have at least 
	250 guests in your park at the end of March, Year 1, with a park rating of at least 600

  correct display should be: 
	在第1年10月末乐园内至少有250游客，并且乐园评分不低于600

---Dynamite Dunes---

  actual display in game: 
	在第1年3月末乐园内至少有
	650游客，并且乐园评分不低于600

  translated from actual display: 
	To have at least 
	650 guests in your park at the end of March, Year 1, with a park rating of at least 600

  correct display should be: 
	在第3年10月末乐园内至少有250游客，并且乐园评分不低于600
	
	
	
![image](https://github.com/OpenRCT2/Localisation/assets/6335952/82659fb8-9386-4d86-bc7e-c3b2a3972cf4)
![image](https://github.com/OpenRCT2/Localisation/assets/6335952/c8a4f63c-02bb-415d-80f6-473c87553cf4)
![image](https://github.com/OpenRCT2/Localisation/assets/6335952/43d318f4-1d9e-4064-be05-6ac374f949f0)

mrmagic2020:
Hi yanleiberg, I had a look into the translation file, and indeed for some reason an extra `{NEWLINE}` is inserted before `{COMMA32}` (where the actual value gets inserted). Removing the `{NEWLINE}` solves the issue. 
[...]

PR #2867, opened 2024-06-28
fix for ISSUE 2866

yanleiberg:
Applying fix for issue(s):
- #2866 

Add extra pair of {POP16}{PUSH16} in STR_2386 for zh_cn, zh_tw, ja-JP.
Removed {NEWLINE} in STR_2386 for zh_cn for better layout.
Add "底" （meaning "the end of" a month) in STR_2386 and STR_2387 to avoid ambiguity in current string, as it isn't currently indicating whether the goal is for the beginning or the end of a month.


Gymnasiast:
Thank you!

PR #2956, opened 2024-10-30
Remove scenario and object override strings

AaronVanGeffen:
Ties in with https://github.com/OpenRCT2/OpenRCT2/pull/23070. These overrides have been replaced with objects. (cf. https://github.com/OpenRCT2/objects/pull/350).


Gymnasiast:
As it turns out, the object overrides have been non-functional since the summer of 2023. Could you remove these as well? In a separate commit, so that we can easily find them back.

Gymnasiast (approved):

tupaschoal:
https://github.com/OpenRCT2/Localisation/blob/master/.github/workflows/translation_check.py can now be simplified

Gymnasiast:
@tupaschoal https://github.com/OpenRCT2/Localisation/pull/2958

PR #2969, opened 2024-11-12
pl-PL: translate scenario names, improve existing translations

arekdurlik:
Another big overhaul, because I've found my old copy of the Triple Thrill Pack in Polish 😁

### DIFF
--- a/objects/pl-PL.json
+++ b/objects/pl-PL.json
@@ -3017,51 +3017,51 @@
     },
     "rct2.scenario_text.build_your_own_six_flags_belgium": {
         "reference-name": "Build your own Six Flags Belgium",
-        "name": "Wybuduj własny park „Sześć Flag” - Belgia",
+        "name": "Wybuduj własny park „Six Flags Belgium”",
         "reference-park_name": "Six Flags Belgium",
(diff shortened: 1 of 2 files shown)


objects/pl-PL.json
```diff
@@ -2993,217 +2993,217 @@
...
-        "park_name": "Six Flags Belgium",
+        "park_name": "„Sześć Flag” - Belgia",
         "reference-details": "Try your hand at running and improving this Six Flags park",
-        "details": "Spróbuj poprowadzić i rozbudować ten park “Six Flags”"
+        "details": "Spróbuj poprowadzić i rozbudować ten park „Sześciu Flag”"
     },
     "rct2.scenario_text.six_flags_great_adventure": {
         "reference-name": "Six Flags Great Adventure",
-        "name": "Six Flags Great Adventure",
+        "name": "„Sześć Flag” - wielka przygoda",
```
Gymnasiast, line 3172: These are all real parks, so you may want to leave these as-is. It’s ultimately up to you to decide, though.

arekdurlik: That's how they were originally translated, but I agree, since I wouldn't personally translate all the other parks that you've listed. I'll revert these changes.

PR #2975, opened 2024-12-05
fix wrong order in heightmap smoothing window

arekdurlik:
The input window for heightmap smooth strength has strings in the wrong order. 1-20 is the range that the buttons to change the value allow, so I guess that's correct. 

Before:
image

After:
image


AaronVanGeffen:
Rather than swapping the strings in the language packs, I suggest we swap the numbers in UiStringIds.h:
```cpp
        STR_ENTER_SMOOTH_STRENGTH = 6708,
// ...
        STR_SMOOTH_STRENGTH = 6709,
```

arekdurlik:
That's a much nicer way of fixing it. Should I close this and open a PR there? I haven't been able to build the game myself though, and won't be able to test it, so maybe an issue would be better? 😅

AaronVanGeffen:
No worries. I've filed https://github.com/OpenRCT2/OpenRCT2/pull/23334 to replace this PR.

PR #2976, opened 2024-12-06
pl-PL: Fix and improve

arekdurlik:
Found some more things to fix. I think the objects file can be left alone for now, at least until the new Wooden Twister is added.

pl-PL.json:
- almost all rollercoasters that had just the word "kolejka" in their name, started with that word, except for three, so I adjusted them for consistency,
- changed the RCT1 Corkscrew name and description to be the same as RCT2,
- capitalized pronouns,
- better name for Dynamite Dunes,
- smaller fixes.

pl-PL.txt:
- mostly small fixes and improved consistency.

### DIFF
--- a/data/language/pl-PL.txt
+++ b/data/language/pl-PL.txt
@@ -4,7 +4,7 @@
 STR_0000    :
 STR_0001    :{STRINGID} {COMMA16}
 STR_0002    :Kolejka spiralna
-STR_0003    :Stojąca kolejka
+STR_0003    :Kolejka stojąca
 STR_0004    :Kolejka podwieszona
 STR_0005    :Kolejka odwrotna
 STR_0006    :Kolejka dla dzieci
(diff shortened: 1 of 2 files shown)


Gymnasiast:
I have just re-dumped the translation, so you can modify the Classic Wooden Twister if desired.

arekdurlik:
Looks good, no need to change. Just the .txt strings left to translate, but I wanted to wait for an issue as to not make a mess.

Gymnasiast:
What do you mean exactly by the .txt changes?

arekdurlik:
The strings STR_0099 and STR_0609. I noticed they were added, but wanted to wait for an issue before adding them to pl-PL.txt

Gymnasiast:
I completely forgot about that. Created an issue here: https://github.com/OpenRCT2/Localisation/issues/2977

PR #2996, opened 2024-12-31
ru-RU: Retranslate objects JSON

andOlga:
Here's the second part of the Russian localisation update.

You may have noticed the enormous size of this patch... unlike the UI strings, which just needed some minor fixes and updates, this one is a complete rewrite of all the text. This is out of necessity: this file is the *most baffling* array of strings I have seen in my entire life. The problems with the current one can be summarized as follows:

- The scenery elements have had all descriptors stripped from them, reduced to base nouns, and reused as much as possible. You want a "Heavenly Bath Roof"? Sorry mate, all I can give you is a "Pool Part". An Atlas Statue? No, you get a "globe"! Many have been reduced to simply "block" or "part" with no other clarification.
- The ride descriptions, likewise, were stripped of all flavor text, and then re-used across all vaguely-similar rides. Feel excited about "locking horns with others in triceratops dodgems", or, perhaps, "battling" them in dragonhead-shaped ones? Nope, you just get to "ride a small electric car".
- Scenario descriptions, while being *mostly* fine, still suffered from some cuts to the text here and there.
- A *lot* of the terms have been misinterpreted and otherwise badly mangled. Here are just *some* of the horrors you've unknowingly been subjecting your Russian peeps to (someone make a Halloween mod out of these):
  - Tents, *everywhere*, instead of a vast variety of different stalls.
    - Plenty of them sell something completely different from what they do in English.
      - One apparently serves alcohol now! (Wine instead of Sujeonggwa).
  - A park on an "unfamous" island, instead of an "infamous" one.
  - "Dudes" instead of passengers (but only on select rides!).
  - "Microbe-uses", instead of micro-buses.
  - "Triceraptors".
  - Carnivorous plants, instead of grapevines.
  - All of the mushrooms you've been planting? Posion. Pure poison, every single one.
  - The joy of going on a "fire race" instead of a "firecracker ride". 
  - Some lads called Gahdes and Manta Ray walking around your park, instead of Hades statues and, you know, the bloody fish.
  - ***Austrians***. Instead of ostriches... because Österreich? I guess???

Having re-translated the file, I've fixed all of these issues (and *many* more), with the text now closely matching the original English one. Strings that are identical in English should now be identical in Russian as well, for easier maintenance. The missing translations were, of course, also added.

With this merged, the Russian translation will be fully up-to-date.

GitHub seems to be struggling with rendering the diff, by the way, so [here](https://github.com/OpenRCT2/Localisation/pull/2996.patch)'s a raw patch file for easier review.

### DIFF
--- a/objects/ru-RU.json
+++ b/objects/ru-RU.json
@@ -805,7 +805,7 @@
     },
     "rct1aa.ride.virginia_reel_tubs": {
         "reference-name": "Virginia Reel Tubs",
-        "name": "Трубы \"Вирджиния\"",
+        "name": "Трубы «Вирджиния»",
         "reference-description": "Circular cars that spin around as they travel along the track",
         "description": "Круглые машины, которые вращаются во время езды.",
         "reference-capacity": "4 passengers per car",
@@ -1149,7 +1149,7 @@
     },
     "rct1ll.ride.face_off_cars": {
         "reference-name": "Face-off Cars",
-        "name": "Машины \"лицом-к-лицу\"",
+        "name": "Машины «лицом-к-лицу»",
         "reference-description": "Riders sit in pairs facing either forwards or backwards as they loop and twist through tight inversions",
         "description": "Люди сидят парами друг напротив друга, проходя резкие петли и перевороты",
         "reference-capacity": "4 passengers per car",
@@ -1513,7 +1513,7 @@
     },
     "rct2.audio.circus": {
         "reference-name": "Ambient sound effects for the Circus attraction.",
-        "name": "Звуковые эффекты для аттракциона \"Цирк\"."
+        "name": "Звуковые эффекты для аттракциона «Цирк»."
     },
     "rct2.audio.title": {
         "reference-name": "Title screen music for RollerCoaster Tycoon 2.",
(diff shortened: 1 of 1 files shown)


AaronVanGeffen:
This must've been _hours_ of work. Thank you very much!

objects/ru-RU.json
```diff
@@ -5603,719 +5603,719 @@
...
         "reference-description": "Self-drive go-karts in Stone Age style",
-        "description": "Люди гоняются на стильных машинках",
+        "description": "Машины для картинга в стиле каменного века",
         "reference-capacity": "Single-seater",
         "capacity": "Одиночные"
     },
     "rct2tt.ride.cerberus": {
         "reference-name": "Cerberus Trains",
-        "name": "Горки Цербер",
+        "name": "Поезда \"Цербер\"",
```
Gymnasiast, line 5754: We always use typographical quotation marks in the OpenRCT2 translation, rather than straight (typewriter) ones. So I guess this would be rendered as `Поезда „Цербер“` or `Поезда «Цербер»`? (I’ll leave it to you to decide the most appropriate one.)

Gymnasiast:
Nice work!

The Russian translation was taken from the GOG release, which itself appears to be taken from the Akella release. Apart from yo-fication, not much has happened to it since.

(That release had more weird things. They extended the sprite font to include the Russian alphabet, but just blacked out the banner font. Took me quite a bit of work to draw up all those characters :') )

objects/ru-RU.json
AaronVanGeffen: It seems the ru-RU.txt file is using typewriter quotes as well, for what it's worth.

andOlga: Replaced with the "christmas tree" variant («these») as they're far more common.

AaronVanGeffen:
@andOlga Please also file a PR against [contributors.md in the main repo](https://github.com/OpenRCT2/OpenRCT2/blob/develop/contributors.md) to add yourself to the contributors to the Russian translation. You deserve the credit! :)

Gymnasiast (approved):

PR #3007, opened 2025-01-09
Sync object translations

AaronVanGeffen:


objects/ru-RU.json
```diff
@@ -6123,113 +6123,113 @@
     },
     "rct2tt.scenario_text.alcatraz": {
         "reference-name": "Alcatraz",
-        "name": "Alcatraz",
+        "name": "Алкатрас",
```
Gymnasiast, line 6126: This looks like it’s unintentionally overwriting newer translations.

@andOlga Which strings are correct here?

AaronVanGeffen: The other way around, surely? The English string was still present in this repository. I take it Russian translations were merged directly into the objects repo?

andOlga: I have intentionally reverted all scenario names to English for several reasons I understand if you want to override the decision, just let me know and I'll translate all of them sometime later, but here's why:

1. It's fairly common for business names to be in English in Russia, including businesses which exist solely within the country and serve only locals. It's common enough in the amusement park sphere as well. This isn't to a point where it's the "one true way" or anything: it's not how things necessarily should be, it's just that something like that doesn't come accross as *wrong*.
2. A LOT of the scenario names are things which sound pleasant in English -- alliterations and the like -- and just kind of... don't... when they're translated literally to Russian. This could be handled by making entirely new park names up, but I'm not sure I can make up stuff that both fits *and* alliterates cleanly for a lot of cases.
3. Consistency - having *some* park names be in English and some (less "pretty" ones, for example) in Russian is weirder than having them all be one way or the other.

AaronVanGeffen: Ah, good that we checked, then. I'll revert the changes in this PR.

Gymnasiast: @andOlga did all her changes to Localisation, and the scenario strings were already present in the dump at the time.

Also, the Dutch translation also very deliberately uses English scenario names, so the presence of English doesn’t mean it was untranslated.

Gymnasiast: Also, we leave these kinds of decision to the translators. After all, it has to make sense in the target language. It’s why I @’ed you in the first place.

Gymnasiast (approved):

Issue #2945, opened 2024-09-30
String 6677–6709


AaronVanGeffen:
Updated strings on PR OpenRCT2/OpenRCT2#22735.

```diff
-STR_2686    :Low:
+STR_2686    :Min. land height:
-STR_2687    :High:
+STR_2687    :Max. land height:
-STR_6045    :Strength
+STR_6045    :Strength:
-STR_6047    :Smooth tiles
+STR_6047    :Smooth tile edges
```

New strings:

```
STR_6677    :Add beaches around water bodies
STR_6678    :Heightmap source:
STR_6679    :Flatland
STR_6680    :Simplex noise
STR_6681    :Heightmap file
STR_6682    :Map Generator - Generator
STR_6683    :Map Generator - Terrain
STR_6684    :Map Generator - Water
STR_6685    :Map Generator - Forests
STR_6686    :Tree to land ratio:
STR_6687    :Min. tree altitude:
STR_6688    :Max. tree altitude:
STR_6689    :{UINT16}%
STR_6690    :Minimum land height
STR_6691    :Enter min. land height between {COMMA16} and {COMMA16}
STR_6692    :Maximum land height
STR_6693    :Enter max. land height between {COMMA16} and {COMMA16}
STR_6694    :Minimum tree altitude
STR_6695    :Enter min. tree altitude between {COMMA16} and {COMMA16}
STR_6696    :Maximum tree altitude
STR_6697    :Enter max. tree altitude between {COMMA16} and {COMMA16}
STR_6698    :Tree to land ratio
STR_6699    :Enter tree to land ratio between {COMMA16} and {COMMA16}
STR_6700    :Simplex Base Frequency
STR_6701    :Enter Base Frequency between {COMMA2DP32} and {COMMA2DP32}
STR_6702    :Simplex Octaves
STR_6703    :Enter Octaves between {COMMA16} and {COMMA16}
STR_6704    :{COMMA2DP32}
STR_6705    :Browse...
STR_6706    :{WINDOW_COLOUR_2}Current image file: {BLACK}{STRING}
STR_6707    :(none selected)
STR_6708    :Smooth Strength
STR_6709    :Enter Smooth Strength between {COMMA16} and {COMMA16}
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
[...]

PR #3032, opened 2025-01-27
es-ES: Apply #2945 + tweaks

dimateos:
Applying for issue(s):
- #2945

Apart from that, I addressed some small issues I came across:

- "cuadrícula": added missing accents
- "Mín/máx": coherency across strings + all accented
- "Clic" usage over "click": IMO "click" is better but ES dictionary has yet to accept it+ there were more strings using "clic"
- Track design / designer: coherency across strings + upper/lower case more similar to EN
- "Paisaje": upper/lower case more similar to EN
- Translated STR_2685 (simplex parameters)

Btw sorry for the absence, it's being some busy months!
Best regards :frog:

### DIFF
--- a/data/language/es-ES.txt
+++ b/data/language/es-ES.txt
@@ -277,11 +277,11 @@ STR_0880    :No se puede elevar el terreno…
 STR_0881    :Un objeto interfiere
 STR_0882    :Cargar Partida
 STR_0883    :Guardar Partida
-STR_0884    :Cargar paisaje
-STR_0885    :Guardar paisaje
-STR_0887    :Salir del editor de escenarios
-STR_0888    :Salir del editor de Vías
-STR_0889    :Salir del admin. de Vías
+STR_0884    :Cargar Paisaje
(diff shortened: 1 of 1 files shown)


Gymnasiast:
> upper/lower case more similar to EN

This is not necessary. The translation has to make sense in Spanish. Literal translations are only good for Greek and Latin studies ;-)

dimateos:
Welp I guess my description could be misleading...

I would not simply use incorrect/uncommon nor a literal translation just to follow EN casing.
In those strings I reference, some words were capitalized to deliver more _meaning_, e.g. a god vs the God.

So apart from making _wording_ consistent I also made _casing_ consistent (easier to contextualize related strings).
I ended capitalizing in a similar way to EN were it felt _right_ and _simpler_ for the user e.g. "the Track Designer" vs "a track design".
This is also consistent with most strings in the ES file that were translated following a similar approach.
Seems like other languages took the same decision e.g. PT :)

PR #3119, opened 2025-05-12
ko-KR: Apply #3120 and etc.

telk5093:
I noticed that `Clear` is used for the button's label of RCT1 path settings.
image

And we prefer `CapsLock` instead of `Caps`.


Gymnasiast:
Hm, the ‘Clear’ thing may just be a bug in that PR. I’ll have to check that.

AaronVanGeffen:
Addressed in https://github.com/OpenRCT2/OpenRCT2/pull/24387. See also https://github.com/OpenRCT2/Localisation/issues/3120

PR #3145, opened 2025-06-10
Remove extraneous formatting codes

AaronVanGeffen:
Ties in with https://github.com/OpenRCT2/OpenRCT2/pull/24559


AaronVanGeffen:
@Gymnasiast With https://github.com/OpenRCT2/OpenRCT2/pull/24559 in, this is ready to be merged as well. Is that alright with you?

Issue #3045, opened 2025-02-19
String 6109


ZeeMaji:
Updated strings on PR OpenRCT2/OpenRCT2#23853.
Worth noting that this is a reintroduction of a vanilla RCT2 string.

```
STR_6109    :Hyper-Twister Roller Coaster    :...
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @ninjum @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @arekdurlik @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken @andOlga

arekdurlik:
Polish translation is fine as is.

groenroos:
`fi-FI` is okay as-is.

Issue #3035, opened 2025-01-30
String 6732


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#23730.

```
STR_6732    :Show a button in the toolbar to rotate the view anti-clockwise
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @ninjum @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @arekdurlik @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken @andOlga

Issue #3024, opened 2025-01-18
Strings 6727-6728


tupaschoal:
New strings from PR OpenRCT2/OpenRCT2#23569.

```
STR_6727    :Dive Loop (left)
STR_6728    :Dive Loop (right)
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [X] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @ninjum @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @arekdurlik @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken @andOlga

Issue #3014, opened 2025-01-10
String 6725, 6726


Gymnasiast:
New strings on PR OpenRCT2/OpenRCT2#23591.

```
STR_6725    :X:
STR_6726    :Y:
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @ninjum @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @arekdurlik @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken @andOlga

andOlga:
...Would these be different in *any* language, ever? I mean I don't mind copying them over but these feel like they don't need to be localizable. Pretty sure math is a universal language there, heh.

PR #3229, opened 2025-07-08
es-ES: Apply #3005, #3009, #3014, #3019, #3024, #3035, #3037, #3045

dimateos:
Hello! long time :)

Applying for issue(s):
- #3005
- #3009
- #3014
- #3019
- #3024
- #3035
- #3037
- #3045

Btw, will probably create a PR to change all "rizo" related words for equivalent "loop"
* I kept it at the beginning for consistency
* But it sounds quite weird, IMO you can see some English words like this in themeparks so it fits
* Other langs keep "loop" similarly e.g. es-CA or pt-BR

Why? (added on edit)
* "rizo" *could* be used for a twisting shape something, but is more commonly used for hair "curls"
* My guess is that "rizo" was picked to try to differentiate "ciclo" "vuelta" "giro", which all could mean "loop"...
  * In fact, "vuelta" would be the most natural, but could be too easly confused with "ciclo" or even "ride"
  * It will always be a compromise!
* In Spanish nowdays you can see a bunch of English words, specially in the media
  * I find it best, to juse use "loop" as in the mentioned languages as example
  * Feels more natural, fits the theme, and there is no compromise

### DIFF
--- a/data/language/es-ES.txt
+++ b/data/language/es-ES.txt
@@ -2166,7 +2166,6 @@ STR_3136    :Advertencia: Este diseño se construirá con un tipo de vehículo a
 STR_3137    :Selecc. Escenario Cercano
 STR_3138    :Reiniciar Selecc.
 STR_3139    :Cable de elevación no puede funcionar en este modo
-STR_3140    :Cable de elevación debe comenzar después de la estación
 STR_3141    :Multi-circuito por attracc. no es posible con un cable de elevación
 STR_3142    :{WINDOW_COLOUR_2}Capacidad: {BLACK}{STRINGID}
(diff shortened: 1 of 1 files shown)


dimateos:
Hey! thanks for the story and advice!

Yep some older strings are less consistent/accurate but over the years I usually search for the words related to new strings and at least try to:

1. Fix related previous spelling
2. Fix related previous wrong
3. Keep consitency with related previous if it seems ok
3. Ensure consintent wording across related strings in the file

I feel like es-ES.txt is quite ok, es-ES.json has quite some missing strings tho!

Happy translating to you too :)

Issue #2960, opened 2024-11-04
String 0098, 0608, 6541


telk5093:
Updated strings on PR OpenRCT2/OpenRCT2#22797.

```diff
-STR_6541    :{WINDOW_COLOUR_2}Rocky Mountain Construction Group, Josef Wiegand GmbH & Co. KG
+STR_6541    :{WINDOW_COLOUR_2}Rocky Mountain Construction Group, Josef Wiegand GmbH & Co. KG, Intamin Amusement Rides Int. Corp. Est.
```

New strings:

```
STR_0098    :LSM Launched Roller Coaster
STR_0608    :Roller coaster trains are accelerated by linear synchronous motors, speeding through tight twists and turns
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [ ] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [ ] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @LPSGizmo @Wuzzy2 @gr33ndev @tellovishous @dimateos @groenroos @rmnvgr @LucaRed @telk5093 @Mmmmulder @Gymnasiast @marcinkunert @Tupaschoal @CsyeCokTheSolly @LPR-Michael @mrmagic2020 @daihakken

PR #3239, opened 2025-07-12
Update sv-SE for #2960

millennIumAMbiguity:
Applying for issue:
- #2960


tupaschoal:
Thanks for all the catching up on the `sv-SE` langs @millennIumAMbiguity !

When doing PRs to comply with open issues, please follow the commit pattern:
```
lang_code: Apply #issue_number
```
Like here:
```
sv-SE: Apply #2960
```

It standardizes our repo and makes it easier to see when and which languages have been changing :)
I fixed all of the PRs that were open to be compliant with this before merging.

PR #3285, opened 2025-07-29
gl-ES: Apply #3268 #3277

ninjum:
Applying for issue(s):
- #3268 
- #3277

PR #3301, opened 2025-08-08
gl-ES: Fix checking of STR_1294

tupaschoal:
Noticed on #3285 that for some reason the script was pointing this as untranslated. No idea what was happening there


tupaschoal:
@ninjum this fixed the remaining one that was claimed to be untranslated but was. I don't see any difference in the characters, let me know if you do.

PR #3302, opened 2025-08-08
Remove unexpected translations from all languages

tupaschoal:
I'm removing all translations reported as "unexpected" because they were not needed. Opening as draft because I want to review if there are more "keys to ignore" that we haven't been maintaining

### DIFF
--- a/data/language/ar-EG.txt
+++ b/data/language/ar-EG.txt
@@ -606,8 +606,6 @@ STR_1214    :Select maximum length of time to wait before departing
 STR_1215    :{WINDOW_COLOUR_2}Synchronise with adjacent stations
 STR_1216    :Select whether to synchronise departure with all adjacent stations (for ‘racing’)
 STR_1217    :{COMMA16}ثواني
-STR_1218    :{BLACK}➕
-STR_1219    :{BLACK}➖
 STR_1220    :مخرج فقط
 STR_1221    :لا مدخل
 STR_1222    :لا مخرج
--- a/data/language/cs-CZ.txt
+++ b/data/language/cs-CZ.txt
@@ -613,8 +613,6 @@ STR_1214    :Vyberte maximální délku čekání před odjezdem
 STR_1215    :{WINDOW_COLOUR_2}Synchronizovat se sousedícími stanicemi
 STR_1216    :Vyberte, zda synchronizovat odjezdy se sousedícími stanicemi (pro „závodění“)
 STR_1217    :{COMMA16} sekund
-STR_1218    :{BLACK}➕
-STR_1219    :{BLACK}➖
 STR_1220    :Jen východ
 STR_1221    :Nemá vchod
 STR_1222    :Chybí východ
@@ -1125,7 +1123,6 @@ STR_1727    :Stavební práva nejsou na prodej!
 STR_1728    :Tyto stavební práva nelze zakoupit…
 STR_1729    :Pozemek není ve vlastnictví parku!
 STR_1730    :{RED}Zavřeno
-STR_1731    :{WHITE}{STRINGID}
 STR_1732    :Stavět
 STR_1733    :Mód
 STR_1734    :{WINDOW_COLOUR_2}Počet kol:
--- a/data/language/da-DK.txt
+++ b/data/language/da-DK.txt
@@ -248,8 +248,6 @@ STR_0831    :Zoom ud
 STR_0832    :Roter synsvinklen 90° med uret
 STR_0833    :Sæt spillet på pause
 STR_0834    :Disk og spil indstillinger
-STR_0839    :{UINT16} x {UINT16}
-STR_0840    :{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{UINT16} x {UINT16}
 STR_0847    :Om ‘OpenRCT2’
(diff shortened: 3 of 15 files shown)


tygrysek90 (commented):
**I have to express my uttermost concern over actually translated and used strings being selected for removal. I want to express my feeling that this is not a good idea.**

At first, it can be observed that some combination of characters (or some utf characters, i mean with higher ordinary values, above ascii) triggers unexpected warnings.

I checked just a few instances of strings presented here for removal and found that they are, in fact, present in en-GB original (and moreover, they are actually used in the game).

For reference, i have assembled a list showing where are individual strings used, put it in localisation wiki: https://github.com/OpenRCT2/Localisation/wiki/What-goes-where (I am sorry, but is so long you have sometimes try refreshing a few times before it loads)

**There are truly lines in translations, that are translating STR-s not present in en-GB.txt, if you would like to clean that, i can submit PR with ease and joy :)**

However, please, strongly reconsider merging this PR, and deleting used strings. please :)

tygrysek90:
I have conducted another research to address the primary question:
 
> (...) I want to review if there are more "keys to ignore" that we haven't been maintaining

Answer is yes. _(But I still insist on noting that those keys might get "translated" for some reason, as I thoroughly analyzed in comments before)_ **I wrote a script to seek strings that don't seem to have anything translatable in them**. Also included functionality to seek against `KEYS_TO_IGNORE` in ` .github/workflows/translation_check.py`. Now i can only present a rough result, but it made worth finds. 

**It states _unknown_ meaning "not present in `KEYS_TO_IGNORE`".**

```
Unknown string with only { } at STR_1142    :{MOVE_X}{10}{STRINGID}
Unknown string with only { } and non text at STR_1143    :»{MOVE_X}{10}{STRINGID}
Unknown string with only { } at STR_1156    :{MOVE_X}{10}{STRINGID}
Unknown string with only { } and non text at STR_1157    :✓{MOVE_X}{10}{STRINGID}
Unknown string with only { } at STR_1736    :{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{POP16}{COMMA16}
Unknown string with only { } at STR_2777    :{MOVE_X}{10}{STRING}
Unknown string with only { } and non text at STR_2778    :»{MOVE_X}{10}{STRING}
Unknown string with only { } and non text at STR_3247    :{COMMA16}%
Unknown string without letters atSTR_5151    :,
Unknown string without letters atSTR_5152    :.
Unknown string with only { } and non text at STR_5813    :“{STRING}”
Unknown string with only { } and non text at STR_5814    :{WINDOW_COLOUR_1}“{STRING}”
Unknown string with only { } and non text at STR_6308    :{TOPAZ}“{STRINGID}{OUTLINE}{TOPAZ}”{NEWLINE}{STRINGID}
Unknown string with only { } and non text at STR_6438    :{MOVE_X}{2}👁
Unknown string with only { } and non text at STR_6450    :{BLACK}“{STRING}”
Unknown string with only { } and non text at STR_6451    :{BLACK}“{STRING}” - {STRING}
Unknown string with only { } at STR_6496    :{WINDOW_COLOUR_2}{STRINGID}
Unknown string with only { } and non text at STR_6642    :{STRING} ({COMMA32} / {COMMA32})
Unknown string with only { } and non text at STR_6651    :{STRING} ({COMMA32}%)
Unknown string with only { } and non text at STR_6689    :{UINT16}%
Unknown string with only { } at STR_6704    :{COMMA2DP32}
Unknown string with only { } and non text at STR_6713    :{COMMA32} {STRINGID}
Unknown string with only { } and non text at STR_6744    :{WINDOW_COLOUR_2}{UINT16}%
Unknown string with only { } at STR_6795    :{MOVE_X}{10}{STRINGID}
Unknown string with only { } and non text at STR_6796    :•{MOVE_X}{10}{STRINGID}
Unknown string with only { } and non text at STR_7002    :{STRINGID} {STRINGID}

```
[check_regulary.py.txt](https://github.com/user-attachments/files/21691357/check_regulary.py.txt) was used to produce this (i had to add .txt suffix because of github restrictions), it needs en-GB.txt from main repo and translation_check.py from this repo, output is best filtered using `| grep Unknown` - still its just very rough and logic could be improved.

For sure this answers question if there are strings that should not require translation, but as i wrote before...

**If you found this output usefull, I could enhance it a little bit and prepare PR to update `KEYS_TO_IGNORE` in `.github/workflows/translation_check.py`**

tupaschoal:
Thanks for looking so thoroughly into this, I highly appreciate it. It is true that I'm removing strings that are in "KEYS_TO_IGNORE" which techinically can be translated, but were not expected to be, since they were in the list to ignore.

I'm ok with reviewing the "KEYS_TO_IGNORE" to remove some or report them differently. My next step was to actually review if there were strings we should be ignoring, but are not, which is what your last answer covers. I'd hapily take a PR to update it with these new additions, then we can later tackle if we want to remove some.

tygrysek90:
> Thanks for looking so thoroughly into this, I highly appreciate it. 

You are welcome, it was my pleasure to do so - you know I like to dig in strings :)

> I'm ok with reviewing the "KEYS_TO_IGNORE" to remove some or report them differently. My next step was to actually review if there were strings we should be ignoring, but are not, which is what your last answer covers. I'd hapily take a PR to update it with these new additions, then we can later tackle if we want to remove some.

Sure, I will bring one PR to add some keys and draft second one on how we could report it in better way. I am still busy with the greenhouse, thus i will do so in evening central European time :)

PR #3303, opened 2025-08-09
Simplify translation_check.py to dump once per language

tupaschoal:
This prints the summary for each language only once, which makes it easier for maintainers to check what they need to do and should greatly reduce the size of the CI job log


tupaschoal:
For reference, here are two runs with and without PR
[log_with_PR.txt](https://github.com/user-attachments/files/21697396/log_with_PR.txt): 290KB, 298 lines
[log_without_PR.txt](https://github.com/user-attachments/files/21697397/log_without_PR.txt): 1058KB, 14256 lines

PR #3304, opened 2025-08-09
Remove unnecessary translations

tupaschoal:
This removes translations to keys that don't exist in `en-GB` anymore for all languages.


tygrysek90 (commented):
Yes, yes, this is very good :)

PR #3309, opened 2025-08-09
Scripting: create possibility to distinguish if technical strings ('unexpected translations') were localized on purpose or altered by mistake

tygrysek90:
This PR:
- changes slightly wording in `translation_check.py` in line for so-called `'Unexpected'` translations, produces warning `Warning: Supossed to remain intact` <- foremost-ly to avoid possible confusion, to be more clear and understandable for future use
 - since it was found that some so-called _technical strings_, as found in `KEYS_TO_IGNORE` are necessary to be localized in a few translations, a **warning suppression mechanism is created**:
    - this expects each language maintainer reviewing issued warning (checking if localization is needed or string was changed by mistake)
    - in case language maintainer finds that strings must be somehow localized, language maintainer will need to commit PR adding respective STR_numbers to newly created `suppress_warning.py`
      - this happens while taking advantage of all existing procedures (PR-s, comments, blame history, etc) which we like to use 

- the PR tackles part of problems defined by discussion with @tupaschoal in
  - https://github.com/OpenRCT2/Localisation/pull/3302
- opening currently as draft - to discuss - if nothing else, there is probably bunch of typos, as I realized i don't have a spellcheck in IDLE 
- thank you for reading


.github/workflows/supress_warning.py
```diff
@@ -0,0 +1,53 @@
...
+#
+# each maintainer kindly PR's STR_nnnn he has checked and found being reported
+# incorectly
+
+
+class DictWithoutKeyError(dict):
+# this is to eliminate KeyError that could occur when acessing
+# language not listed here
+    def __missing__(self, key):
+        return []
```
tupaschoal, line 12: Do we need this? We should have keys for all languages, right? Letting the error in will remind us to add the new key here when we add a new language.

.github/workflows/translation_check.py
```diff
@@ -148,7 +150,7 @@ def count_translations(dir_with_translations, print_info, reference_file):
             print(f'\t{', '.join(messages["missing"])}')
             print(f'Unnecessary (not in en-GB) ({not_needed[lang]}):')
             print(f'\t{', '.join(messages["unnecessary"])}')
-            print(f'Unexpected (should not be translated) ({len(messages["unexpected"])}):')
+            print(f'Warning: Supossed to remain intact ({len(messages["unexpected"])})\nPlease review and if OK, add to supress_warning.py:')
```
tupaschoal, line 153:
```suggestion
            print(f'Translation not expected ({len(messages["unexpected"])})\nPlease review and if OK, add to supress_warning.py:')
```

Not 100% satisfied with neither wording yet.

.github/workflows/supress_warning.py
```diff
@@ -0,0 +1,53 @@
...
+
+SUPRESS_WARNING = DictWithoutKeyError()
+
+# example
+# SUPRESS_WARNING['qq-JJ'] = ['STR_0123', 'STR_5678', 'STR_8900']
+# 2996 - 3009 are COLOR-ABC strings used in Banner window
+# japanian and ukrainian use different scripts (hentai and cyrilic)
```
tupaschoal, line 19:
```suggestion
# japanese and ukrainian use different scripts (hentai and cyrillic)
```
What, it is not really hentai, is it?

tupaschoal (changes requested):
I like this idea, of course we should first make sure there are no current wrong "keys_to_ignore", to clean up from the main list, then we can add exclusions here.

.github/workflows/supress_warning.py
tygrysek90: Okay, I was not really sure if you would like to prefer "hiding problems" or "getting problems into face" approach, metaphoricaly said. 
Do you think we need to try to catch possibly created exception so it gets printed in output or leave it without catching and just let script break in case of new language?

.github/workflows/supress_warning.py
tygrysek90: You reviewed well and in detail :) I couldn't remember the word when writing this late night. It is hiragana: https://en.wikipedia.org/wiki/Hiragana 

To my knowledge _hentai_ actually means 'lewd' in japanese

.github/workflows/supress_warning.py
tupaschoal: I'd rather have this blow-up, it's very unlikely for us to be looking at the logs of this CI already and this would fail silently.

tygrysek90:
At this point, suggested and requested changes were incorporated, typos fixed (spelling of the word suppress...)

Only point remaining that is not fine remains:

`            print(f'Translation not expected ({len(messages["unexpected"])})\nPlease review and if OK, add to suppress_warning.py:')`

Maybe could we call it **_technical string_**? And have the message _"Technical string modified in translation"_?

What do you think @tupaschoal ?

tygrysek90:
> It's a bit lengthy, but I think it's alright (we might want to break into two or three print/lines to make more readable on the actual script.

For me personally, it is better to receive longer, but clearer message than to get stub of sentence that will keep me puzzled for next week ;)

For sake of well-reading formatted to 80 column width:

```
Technical strings modified (XX).\n
It's not expected for these to be translated, but if it makes sense for your
language, add a waiver for them on suppress_warning.py and open a PR explaining
why.
```

I will update the PR soon with this version

tygrysek90:
Found that 80 column width is too narrow, hence I chosen to break it in two so it is easier to read and looks logical -
one line is as cause and second as solution

I fined tuned it with commit above, adding condition for the longer (explanation) part of text to be printed only in case there are some "Technical strings modified"

Here is screenshot of output in staged repository:
Snímek obrazovky z 2025-08-11 13-05-40

I feel the PR is ready now.

(We could add Technical strings to comment output, just for one language at time but I would leave that to another time)

.github/workflows/translation_check.py
```diff
@@ -148,8 +150,11 @@ def count_translations(dir_with_translations, print_info, reference_file):
             print(f'\t{', '.join(messages["missing"])}')
             print(f'Unnecessary (not in en-GB) ({not_needed[lang]}):')
             print(f'\t{', '.join(messages["unnecessary"])}')
-            print(f'Unexpected (should not be translated) ({len(messages["unexpected"])}):')
+            print(f'Technical strings modified ({len(messages["unexpected"])}):')
             print(f'\t{', '.join(messages["unexpected"])}')
+            if len(messages["unexpected"]):
```
tupaschoal, line 155: Could be just:
```suggestion
            if messages["unexpected"]:
```
No?

.github/workflows/translation_check.py
```diff
@@ -148,8 +150,11 @@ def count_translations(dir_with_translations, print_info, reference_file):
...
+                print(f'\tIt\'s not expected for these to be translated, but if it makes sense for your language,')
+                print(f'\tadd a waiver for them on suppress_warning.py and open a PR explaining why.')
```
tupaschoal, line 157: I wonder whether it will be confusing if "these" is referring to the strings above or below
```suggestion
                print(f'\tIt\'s not expected for the above to be translated, but if it makes sense for your language,')
                print(f'\tadd a waiver for them on suppress_warning.py and open a PR explaining why.')
```

tupaschoal (commented):
Are the suppressions final or do you intend to patch other languages too?

.github/workflows/translation_check.py
tygrysek90: 🐍👍 

Yes, that is more pythonic, thanks!

tygrysek90:
> Are the suppressions final or do you intend to patch other languages too?

**Well that is a good question!** I have filled the given list with something (so it makes some sense while write-test-review process).

However now I feel **I should actually purge the suppression list.** 

For sake of cleanliness and serving complete process:
1. language maintainer observes script output (or repo maintainer observes and creates issue for lang maintainer)
2. language maintainer decides what action to take
3. language maintainer submits a PR resolving situation - _a:_ reverts translation, _b:_ suppresses warning

**Here now, we (together with @tupaschoal ) are standing as script writers. We give a tool** - none of us can not say whether it is right or wrong to have it translated (there might be reasons it is good or bad beyond our understanding, beyond facts we can see from our perspectives)

_To metaphor, when you buy impact drill, the shop assistant will not tell you where you should drill :)_

tupaschoal (approved):

PR #3311, opened 2025-08-10
Scripting: Action to comment PR-s touching scripts

tygrysek90:
Auto-comment PR-s touching scripts with explained diff showing changes to KEYS_TO_IGNORE in translation_check.py

______

Since I found myself helping @tupaschoal with summer cleaning :) of this repository in
- https://github.com/OpenRCT2/Localisation/pull/3302
- https://github.com/OpenRCT2/Localisation/pull/3308

A need to observe what is actually hidden in KEYS_TO_IGNORE and - what possible changes are making - the need araised - since it is quite dull to check 100 indexes by hand. Therefore I made script that does alike locally (thus still requires truckload of copy-pasting) - then I thought it could be useful to set it online.

PR as draft, to review idea and ... I am certain I made some mistake/oversight :)
_______

Screenshot to give idea (made with staged files in staged repository for testing, its just showcase)

staged

______

### DIFF
--- a/.github/workflows/keys-change-check.yml
+++ b/.github/workflows/keys-change-check.yml
@@ -1,0 +1,93 @@
+name: Keys change check
+
+on: 
+  pull_request_target:
+    paths:
+     - '.github/workflows/**.py'
+
+
+jobs:
+  keys-change-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name:  Get Pull Request
+        uses: actions/checkout@v4
+        with:
+          path: pr
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+
+      - name: Get OpenRCT2 develop (for en-GB.txt)
+        uses: actions/checkout@v4
+        with:
+          repository: OpenRCT2/OpenRCT2
+          ref: develop
+          path: OpenRCT2
+
+      - name: Checkout translations before PR changes (for comparsion)
+        uses: actions/checkout@v4
+        with:
+          path: master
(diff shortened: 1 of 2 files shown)


tygrysek90:
@tupaschoal Hi!
Could you say what do you think about this idea?
Is it worth pursuing and completing - including auto-expanded diffs for suppress_warning.py, or is this an excessive and overkill?
I could complete it with ease, but I would really like to hear if it is good, or should I let it go?
Thanks in advance :)

tupaschoal:
I'm not sure I get all the ideas you're trying do here, would you mind explaining it in another way?

tygrysek90:
> I'm not sure I get all the ideas you're trying do here, would you mind explaining it in another way?

@tupaschoal yes - i see that my style of writing PR texts is terrible (i re-read it after myself after two weeks and found i tend to divert from actual topic in writing ... which i am doing once again... *facepalm*)

**The idea is** to create an Action (image) script that would:

- get activated only in case when PR in this repository proposes modification to `translation_check.py` (or `suppress_warning.py`)
- produce output into comment under such PR
  - comment would contain diff-like summary of what is being done
- the envisioned **usefulness of such action is expanding STR_nnnn variables to their full 'meaning'** - so the human person reviewing the PR doesn't have to do any extra look-up
- the secondary envisioned usefulness is to mark STR numbers that are no more present in en-GB.txt, hence the maintainers and reviewers can take according actions

tupaschoal:
I see, thanks for explaining @tygrysek90 . For the removal of strings I don't think it is very important, as we often do it ourselves and if we're searching and removing, it should find on the python script too.
As for helping with the lookup, it might be a good time saver. I'll let @Gymnasiast weigh in

PR #3394, opened 2025-11-30
de-DE: Apply #3383

Wuzzy2:
Applying for issue:
- #3383


Wuzzy2:
Oops, it was indeed a typo. Fixed.

PR #3432, opened 2026-03-05
hi-IN: Base file for hindi localisation

akshayadav:
Base file for the localization support for hindi language


tupaschoal (changes requested):
Hi @akshayadav !

This is quite a work, can you walk us through how you approached translating them all? There have been a few new strings in the meantime too, if you want to complete it.

Additionally, if you can rebase so that the CI re-runs it would be great.

Let us know if you need help.

PR #3456, opened 2026-04-03
eo-ZZ: removal of title case, numerous fixes

slashdevslashurandom:
This rather big pull request does the following things:

# Title case removal

Title case (a convention in which every word of a title is capitalized, sometimes with exception of small particles) is exclusive to English. Its use in Esperanto seems to be much more restricted, in line with other dual-case European languages. Hence, most of its instances have been replaced with sentence case (with the notable exception of the "Wild Mouse" rides, where "Sovaĝa Muso" is treated like a proper name for clarity, as while it does use mouse-shaped vehicles, these aren't the only ones available).

# Tweaks to specific strings

Wild Mouse (0011, 0056): "muŝo" means "fly", "muso" means "mouse".

Observation Tower (0016): "Turo por Observadi" replaced with "Observa turo", the more common name for the real-life object (the former sounds more like "tower for observation").

Go-Karts (0024): "Gokartoj", a loanword, replaced with "Vetkuraj aŭtetoj" ("Small race cars"), as that is the more common term used for karting in Esperanto.

Dodgems / Bumper Cars (0027): "Doĝemoj", a loanword based on an old British trademark, replaced with "Kunfrapemaj aŭtetoj" ("Small cars with an inclination towards collisions"), aligning more with the American English name, the French name and the actual behavior of the ride.

Top Spin (0042): "Tuniĝanta gondolo", a possible typo, replaced with "Turniĝanta gondolo" ("Turning gondola")

# Tweaks to grammar

Multi-Dimension Roller Coaster (0057): more natural phrasing.

'S' Bend (0929-0930), 'U' shaped open track (1460), 'O' shaped enclosed track (1461): previous grammar implied that "shape" is the noun, rather than "track".

Save this before...? (0947-0949): Questions with "yes/no"-style answers are supposed to have a "ĉu" particle at the start.

OpenRCT2 file types (1043-1046): More natural word order.

Bumper-car mode (1074): Changed to match the new ride name in 0027.

Entry/Exit platform (1409): More natural phrasing that, this time, actually matches English more.

SHIFT +, CTRL + (2782-2783): Replaced labels with accepted Esperanto names for the corresponding keys ("MAJ" short for "majuskla", "capital (letter)", "STIR" short for "Stir-klavo", "Control key")

RCT1 widgets (5282-5284, 6141): Clarified with "RCT1-style" for a more natural sentence.

Setting will take effect after restarting (5558): replaced with "on the next launch of OpenRCT2" for a more natural sentence.

[...]

PR #3457, opened 2026-04-04
eo-ZZ: removal of Title Case and rewording for objects

slashdevslashurandom:
Same process as for the data file, now applied to the objects file for Esperanto.

Title Case removed in names of ride vehicles, objects, etc., with exception of proper names and names of scenarios/parks.

"Bumbly Beach" and "Bumbly Bazaar" had the adjective "burda" (which has a literal sense of "related to a bumblebee" replaced with "plumpa", which, I think, is close enough in meaning to "bumbling", "clumsy" without being too insulting.

References to rides renamed in #3456 updated (surprisingly, one of the lines already uses the same term "aŭtetoj" for dodgems / bumper cars).

Lots of dual-noun scenario/park names (ex. "Diamond Heights") changed to have the first word in its adjective form.

In "Forest Frontiers", the word for "frontiers" changed from "limoj" (which literally means "border" and doesn't cover the meaning of "land untouched by civilization") to "krudlandoj" ("raw/crude lands")

All references to Ancient Rome (scenarios, theming, etc.) changed to use the word "Romio" and its forms, specifically referring to Rome (the historic state), as opposed to Rome (the city).

"Club-shaped Tree" reworded to make sure it references clubs (the playing card suit) instead of clubs (weapons), likewise for "Heart-shaped tree".

Some tree species names changed to ones matching Esperanto Wikipedia.

"Morta" replaced with "senviva" in "dead tree" objects, as the former is supposed to mean "deadly" or "related to the act of dying" (i.e. changing state from alive to dead) as opposed to "dead" (lifeless).

Some uses of phrases like "1920s" and "1960s" clarified with the word "years". (I was actually considering changing the phrase "Roaring Twenties" from a literal translation ("Muĝadaj Dudekaj Jaroj") to a more blunt "Usonaj 1920aj jaroj" (lit.: U.S. American 1920th years), but decided that this could be made as a separate PR with its own discussion later.

"Crater Carnage" changed from literal "Kratera Buĉado" (lit. "Crater Butchery") to "Kratera Katastrofo" to avoid an overly harsh word and restore alliteration.

In "Lost City Founder", changed "Perdita Urbo" (literally "lost city") to "Malaperinta Urbo" ("city which disappeared") to match the Esperanto Wikipedia term and names of lost city theming objects already in the game.

Slight grammar changes in other objects and scenario descriptions.

CC: @tellovishous

### DIFF
--- a/data/language/eo-ZZ.txt
+++ b/data/language/eo-ZZ.txt
@@ -86,8 +86,8 @@ STR_0089    :Miniatura onda fervojo
 STR_0090    :Minrajdo
 STR_0092    :Onda fervojo lanĉita per linearaj induktaj motoroj
 STR_0093    :Hibrida onda fervojo
-STR_0094    :Onda fervojo kun unuopa relo
-STR_0095    :Alpa onda fervojo
+STR_0094    :Unurela onda fervojo
+STR_0095    :Alpeca onda fervojo
 STR_0096    :Klasika ligna onda fervojo
 STR_0097    :Klasika staranta onda fervojo
 STR_0098    :Onda fervojo lanĉita per linearaj sinkronaj motoroj
--- a/objects/eo-ZZ.json
+++ b/objects/eo-ZZ.json
@@ -4313,7 +4313,7 @@
     },
     "rct2.scenery_small.tdt1": {
         "reference-name": "Dead Tree",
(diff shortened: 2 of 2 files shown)


tellovishous:
All of these also look good to me
Dankon denove 👍

PR #3458, opened 2026-04-05
eo-ZZ: reword "Loading..."-style messages into verbal nouns

slashdevslashurandom:
A quite small PR this time. This one rewords messages indicating an ongoing action from a present-tense verb into a verbal noun.

The goal is to make the translation a bit more formal. Instead of having the program "speak for itself" with an implied first person pronoun, the ongoing action is described as a noun. From my experience, this matches with how other European languages handle such strings.

For instance, when it comes to Microsoft software (which, at least until the 2010s, seems to have maintained a high standard for localization, at least in my experience), phrases that start with "Downloading..." in English are generally translated using "Загрузка..." and "Téléchargement..." in Russian and French, using their respective verbal nouns, whereas in German the action is instead written in passive voice with "...(wird/werden) heruntergeladen" ("...(is/are) being downloaded").

Likewise, [Firefox's Esperanto localization](https://pontoon.mozilla.org/eo/firefox/all-resources/?search=downloading) seems to mostly use either the Russian/French approach ("Elŝuto..."/"Elŝutado...") or the German one ("...estas elŝutata").

cc @tellovishous


tellovishous:
Seems reasonable enough, looks good to me 👍

Gymnasiast:
FWIW, Dutch translations tend to use the infinitive "Downloaden...", sometimes prefixed with "Bezig met" (which means something like "working on/busy with".

LucaRed:
Adding more context for future reference.

Italian translations also usually adopt a non-personal voice. E.g.: "Salvataggio in corso..." (literally "*salvation ongoing...*" for "Saving...").

A big exception is macOS, where the computer speaks in first person ("Salvo il documento...", literally "I'm saving the document..."). But, honestly, it sounds weird.

Issue #3461, opened 2026-06-02
String 7030


telk5093:
New strings from PR OpenRCT2/OpenRCT2#26327.

```
STR_7030    :{WINDOW_COLOUR_2}Guests entertained: {BLACK}{COMMA32}
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [x] de-DE
- [x] en-US
- [x] eo-ZZ
- [x] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] gl-ES
- [x] hu-HU
- [x] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [x] pl-PL
- [x] pt-BR
- [x] ru-RU
- [x] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

tupaschoal:
Thanks for filing it @telk5093 !

PR #3462, opened 2026-06-02
ko-KR: Apply #3461 and `STR_0896`

telk5093:
Applying for issue(s):
- #3461

I was shocked that this string has not been issued since last month 😓


tupaschoal:
> I was shocked that this string has not been issued since last month 😓

Sadly it is not uncommon for us to forget it :(

Btw how did you notice `STR_0896`? Was there any side effect? Seems like it is missing an update on most languages, if you don't mind doing a PR to fix them all

telk5093:
> Btw how did you notice STR_0896? Was there any side effect? Seems like it is missing an update on most languages, if you don't mind doing a PR to fix them all

Some user reported me that there is no ride name in the construction window title so I figured out it.
\+ I was going to it after this PR accepted ;)

PR #3463, opened 2026-06-02
ru-RU: Apply #3461

andOlga:
Applying for issue(s):
- #3461


tupaschoal:
@andOlga note that `ru-RU` seems to be missing #3358

Issue #3471, opened 2026-06-09
Strings 7031–7032


AaronVanGeffen:
New strings from PR OpenRCT2/OpenRCT2#26442.

```
STR_7031    :Disable grass growing
STR_7032    :Disable grass ageing so that it doesn't grow.
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [x] de-DE
- [x] en-US
- [x] eo-ZZ
- [x] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] gl-ES
- [x] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [x] pl-PL
- [x] pt-BR
- [x] ru-RU
- [x] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

Issue #3408, opened 2026-01-26
String 7013-7017


Gymnasiast:
Used in the Footpaths window (OpenRCT2/OpenRCT2#25675):
```
STR_7013    :Drag areas of path
```

Used in the first-time user setup (OpenRCT2/OpenRCT2#25643 and OpenRCT2/OpenRCT2#25858):
```
STR_7014    :I own the game on Steam, but I haven’t installed it yet.
STR_7015    :Please close Steam if it’s running, then click ‘OK’.
STR_7016    :OpenRCT2 has tried to trigger a download in Steam. Please open Steam and let it download the game. When Steam is finished, click ‘OK’.
STR_7017    :Quit
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [x] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] gl-ES
- [x] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [x] pt-BR
- [x] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

Issue #3398, opened 2025-12-12
String 7009-7011


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#25575 and OpenRCT2/OpenRCT2#25124.

```
STR_7009    :Receiving scripts…
STR_7010    :Could not start replay, file ‘{STRING}’ doesn’t exist or isn’t valid
STR_7011    :Could not start replay
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

Issue #3395, opened 2025-12-12
String 7012


Patrxgt:
New strings from PR OpenRCT2/OpenRCT2#25621.

```
STR_7012    :Polish Złoty (PLN)
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [x] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] gl-ES
- [x] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [x] pl-PL
- [x] pt-BR
- [x] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

PR #3483, opened 2026-07-06
es-ES: add new translations for attraction types and related messages (STR_7007-STR_7032)

Davis-3450:
I've added missing translations from said ranges

### DIFF
--- a/data/language/es-ES.txt
+++ b/data/language/es-ES.txt
@@ -3804,3 +3804,29 @@ STR_7003    :El archivo de audio ‘{STRING}’ está truncado. Se esperaban {IN
 STR_7004    :Forzar redibujado
 STR_7005    :Arrastrar un área de sendero
 STR_7006    :Dibujar borde alrededor de los botones con imágenes
+STR_7007    :Tipo de atracción
+STR_7008    :Tipo de atracción desconocido ({INT32})
+STR_7009    :Recibiendo scripts…
+STR_7010    :No se pudo iniciar la repetición, el archivo ‘{STRING}’ no existe o no es válido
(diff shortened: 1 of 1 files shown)


Gymnasiast:
Can you list the issues that these strings come from? Here is a list: https://github.com/OpenRCT2/Localisation/issues

(Makes it easier for us to keep track of how much of the file is translated.)

Also, I can list you as the maintainer for Spanish and make sure you get notifications when we add or change strings. Would you like that?

Davis-3450:
Hey! here are the issues:

https://github.com/OpenRCT2/Localisation/issues/3364 https://github.com/OpenRCT2/Localisation/issues/3373 https://github.com/OpenRCT2/Localisation/issues/3383 https://github.com/OpenRCT2/Localisation/issues/3398 https://github.com/OpenRCT2/Localisation/issues/3395 https://github.com/OpenRCT2/Localisation/issues/3408 https://github.com/OpenRCT2/Localisation/issues/3445 https://github.com/OpenRCT2/Localisation/issues/3441 https://github.com/OpenRCT2/Localisation/issues/3461 https://github.com/OpenRCT2/Localisation/issues/3471

And sure I'd love to keep contributing, so yes list me as a maintainer!

I think it'd be useful to add a script to compare lines across languages to
[en-GB.txt](https://github.com/OpenRCT2/OpenRCT2/blob/develop/data/language/en-GB.txt) (which is the source of truth) so we can easily see the coverages per language, missing strings, percentage translated.

kind of like the existing ones but more complete.

Could be implemented in CI or scripts, if you think that's worth it I can build it too!

PR #3490, opened 2026-07-09
Fix: use curl instead of checkout for en-GB.txt

Davis-3450:
Replaces the `actions/checkout` of the `OpenRCT2/OpenRCT2` repo with a direct `curl` download of `en-GB.txt`.

- fetches only the single `en-GB.txt` file instead of cloning the whole repo.
- produces the same path the script expects (`OpenRCT2/data/language/en-GB.txt`) from the same source (`develop` branch).
- doesn't touch permissions, triggers, or the rest of the pipeline.

### DIFF
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -20,7 +20,7 @@ jobs:
       - name: get en-GB.txt (curl)
         run: |
           mkdir -p OpenRCT2/data/language
-          curl -sSL https://raw.githubusercontent.com/OpenRCT2/OpenRCT2/develop/data/language/en-GB.txt -o OpenRCT2/data/language/en-GB.txt
+          curl -sSLf https://raw.githubusercontent.com/OpenRCT2/OpenRCT2/develop/data/language/en-GB.txt -o OpenRCT2/data/language/en-GB.txt
 
       - name: Checkout translations before PR changes (for comparsion)
         uses: actions/checkout@v4
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -17,12 +17,17 @@ jobs:
           ref: ${{github.event.pull_request.head.ref}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
 
-      - name: Get OpenRCT2 develop (for en-GB.txt)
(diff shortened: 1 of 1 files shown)


.github/workflows/translation_check.yml
```diff
@@ -17,12 +17,10 @@ jobs:
...
-      - name: Get OpenRCT2 develop (for en-GB.txt)
-        uses: actions/checkout@v4
-        with:
-          repository: OpenRCT2/OpenRCT2
-          ref: develop
-          path: OpenRCT2
+      - name: get en-GB.txt (curl)
+        run: |
+          mkdir -p OpenRCT2/data/language
+          curl -sSLf https://raw.githubusercontent.com/OpenRCT2/OpenRCT2/develop/data/language/en-GB.txt -o OpenRCT2/data/language/en-GB.txt
```
tupaschoal, line 23: I understand the motivation for not checking out everything so that it can be faster, but I'm not sure using curl is the best option, since we need to work with GH's URL. GitHub actions provides a way to checkout [only a given file](https://github.com/actions/checkout#fetch-only-a-single-file), so we could instead do:
```suggestion
      - name: Get OpenRCT2 develop (for en-GB.txt)
        uses: actions/checkout@v4
        with:
          repository: OpenRCT2/OpenRCT2
          ref: develop
          path: OpenRCT2
          sparse-checkout: |
              data/language/en-GB.txt
          sparse-checkout-cone-mode: false
```

Gymnasiast: GitHub’s URLs haven’t changed much over the years, so I’m comfortable doing it like this. As such, I’ll merge it, but we can revisit it if necessary.

tupaschoal: Why not my suggestion? Seemed more "native"

Davis-3450: Well if you consider the size of the main repo, it can make an important difference. 

If it doesn't work we can always revert it.

tupaschoal: I understand, but so does my suggestion, it only checks out a single file doesn't it?

Davis-3450: I think it cloned the entire repo.

PR #3505, opened 2026-08-02
pt-BR: Apply #3501

tupaschoal:
Applying for issue(s):
- #3501

PR #3506, opened 2026-08-02
Move actions on CI to latest

tupaschoal:
According to CI on #3505

Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/translation_check.yml
+++ b/.github/workflows/translation_check.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name:  Get Pull Request
-        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/translation_check.yml
Gymnasiast:
```suggestion
  pull_request_target:
```

Gymnasiast (approved):

Issue #3435, opened 2026-03-12
rct1.ride.spinning_cars


Gymnasiast:
We just added this new train type to the object translation file.

```
    },
    "rct1.ride.spinning_cars": {
        "reference-name": "Spinning Cars",
        "name": "Spinning Cars",
        "reference-description": "Circular roller coaster cars which are free to spin around their central axis.",
        "description": "Circular roller coaster cars which are free to spin around their central axis.",
        "reference-capacity": "4 passengers per car",
        "capacity": "4 passengers per car"
    },
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [ ] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [ ] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [ ] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

Issue #3355, opened 2025-10-11
Strings 1080, 1081, 1085-1087


matheusvb3:
Updated strings on PR https://github.com/OpenRCT2/OpenRCT2/pull/25332.

```
STR_1080 :Film: “Avenging Aviators”
STR_1081 :3D film: “Mouse Tails”
STR_1085 :Film: “Thrill Riders”
STR_1086 :3D film: “Storm Chasers”
STR_1087 :3D film: “Space Raiders”
```

Updated in:
- [ ] ar-EG
- [x] ca-ES
- [x] cs-CZ
- [x] da-DK
- [x] de-DE
- [ ] en-US
- [x] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [x] fr-FR
- [x] gl-ES
- [x] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [x] pl-PL
- [x] pt-BR
- [x] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

LucaRed:
`it-IT` is OK

PR #3513, opened 2026-08-07
Resolves for es-ES

Davis-3450:
- **#3355** — STR_1080–1081 and STR_1085–1087: opening quote fixed from `”` to `“`, matching en-GB. Text unchanged.
- **#3435** — `rct1.ride.spinning_cars` in `objects/es-ES.json`: `name` and `description` translated. No `reference-*` field touched.

The rest of the open issues were checked against the current en-GB and are already applied in es-ES; both files have the same IDs and keys as en-GB.

Note: #3441 was already ticked for es-ES but the change had never been applied.


Davis-3450:
Just made a lot of fixes!

Issue #3540, opened 2026-09-12
String 7050-7063


Gymnasiast:
New strings from PR OpenRCT2/OpenRCT2#27061.

These strings are used in the Guest List window.

```
STR_7050    :Guests walking
STR_7051    :Guests heading for {STRINGID}
STR_7052    :Guests queuing for {STRINGID}
STR_7053    :Guests drowning
STR_7054    :Guests on {STRINGID}
STR_7055    :Guests in {STRINGID}
STR_7056    :Guests at {STRINGID}
STR_7057    :Guests sitting
STR_7058    :Guests picked up
STR_7059    :Guests watching {STRINGID}
STR_7060    :Guests watching construction of {STRINGID}
STR_7061    :Guests looking at scenery
STR_7062    :Guests leaving the park
STR_7063    :Guests watching new ride being constructed
```

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [x] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [x] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @Davis-3450 @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

Issue #3539, opened 2026-09-12
String 7044–7049


AaronVanGeffen:
New strings from PR OpenRCT2/OpenRCT2#26947.

```
STR_7044    :News ticker
STR_7045    :RCT1 style status bar
STR_7046    :Status bar (RCT1 style)
STR_7047    :Step control (Track Designer)
STR_7048    :Step control (Scenario Editor)
STR_7049    :OpenRCT2
```

Notes:
- The 'status bar' was formerly known as the 'bottom toolbar', but has been split into smaller windows. See also #3514 
- `STR_7045` represents a checkbox in the themes settings tab
- `STR_7049` is used as the caption on the 'RCT1 style' game status bar
- The other strings are windows labels used in the themes window

Updated in:
- [ ] ar-EG
- [ ] ca-ES
- [ ] cs-CZ
- [ ] da-DK
- [ ] de-DE
- [x] en-US
- [x] eo-ZZ
- [ ] es-ES
- [ ] fi-FI
- [ ] fr-FR
- [ ] gl-ES
- [ ] hu-HU
- [ ] it-IT
- [ ] ja-JP
- [x] ko-KR
- [ ] nb-NO
- [x] nl-NL
- [ ] pl-PL
- [ ] pt-BR
- [x] ru-RU
- [ ] sv-SE
- [ ] tr-TR
- [ ] uk-UA
- [ ] vi-VN
- [ ] zh-CN
- [ ] zh-TW

Notification when new strings arrive:
@OmarAglan @J0anJosep @tygrysek90 @LPSGizmo @Wuzzy2 @tellovishous @dimateos @Davis-3450 @rmnvgr @ninjum @LucaRed @telk5093 @Gymnasiast @arekdurlik @Tupaschoal @andOlga @mrmagic2020 @daihakken

PR #3541, opened 2026-09-12
ru-RU: Apply 3539, 3540

andOlga:
I would like to point out that the word "at" is very awkward to translate to Russian because there is no exact equivalent and it's heavily context-dependent which word you'd choose. Without further context, I have chosen to interpret it roughly as "Guests *near* {STRINGID}", but let me know if that needs correction.

Applying for issue(s):
- #3539
- #3540


Gymnasiast:
"Guest at X" is always used to describe guests who are at the counter of a shop (food/drink/souvenir). It’s not used for anything else. Hope this helps.

andOlga:
> counter of a shop

If I recall correctly all the shops in RCT2 are outdoorsy ones that you just approach the counter of from the outer side. In this case the current translation actually happens to be correct here 😄 And if there *are* indoors ones, then I could adjust it, though the result would be somewhat longer (literally like "Guests in the store {STRINGID}").

Gymnasiast:
Shops are indeed all of the outdoorsy type. Guests will only enter facilities like the toilets and first aid rooms, but those use the "Guests in <X>" string instead.

andOlga:
Perfect, can be merged as-is then.

